PR #5, opened 2020-06-02
Node title block

stephen-cox:
Port of the node title block from bhcc_helper.

### DIFF
--- a/config/optional/block.block.nodeheaderblock.yml
+++ b/config/optional/block.block.nodeheaderblock.yml
@@ -1,0 +1,19 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - localgov_core
+  theme:
+    - localgov_theme
+id: nodeheaderblock
+theme: localgov_theme
+region: header
+weight: 0
+provider: null
+plugin: localgov_node_header_block
+settings:
+  id: localgov_node_header_block
+  label: 'Node header block'
+  provider: localgov_core
+  label_display: '0'
+visibility: {  }
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * @file
+ * LocalGovDrupal Core module file.
+ */
+
+/**
+ * Implements hook_theme().
+ */
+function localgov_core_theme() {
+  return [
+    'page_header' => [
+      'variables' => [
+        'title' => '',
+        'lede' => NULL,
+      ],
+      'render element' => 'block',
+    ],
+  ];
+}
--- a/src/Plugin/Block/NodeHeaderBlock.php
+++ b/src/Plugin/Block/NodeHeaderBlock.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace Drupal\localgov_core\Plugin\Block;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Cache\Cache;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Routing\CurrentRouteMatch;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\node\NodeInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'NodeHeaderBlock' block.
+ *
+ * @Block(
+ *  id = "localgov_node_header_block",
+ *  admin_label = @Translation("Node header block"),
+ * )
+ */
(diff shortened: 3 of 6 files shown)


localgov_core.module
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Implements hook_theme().
+ */
+function localgov_core_theme() {
+  return [
+    'page_header' => [
+      'variables' => [
+        'title' => '',
+        'lede' => NULL,
```
Adnan-cds, line 16: Good to see that the curse of "lede" has stuck with us across repositories :D

tests/src/Functional/BlockTest.php
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Test blocks display.
+   */
+  public function testBlocksDisplay() {
+    $this->drupalLogin($this->adminUser);
+
+    // Check node title and summary display on a landing page.
+    $title = $this->randomMachineName(8);
+    $summary = $this->randomMachineName(16);
+    $body = $this->randomMachineName(32);
```
Adnan-cds, line 53: This is fine.  An alternative could be Drupal\Component\Utility\Random::sentences()

config/optional/block.block.nodeheaderblock.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - localgov_core
+  theme:
+    - localgov_theme
```
Adnan-cds, line 7: I wonder if localgov_theme needs to go into the info file and composer.json of this module.  Something to think about.

andybroomfield, line 7: Do we need to have a theme dependency for this block?
Is the plan that this block is only for use with the localgov theme or can it be used outside?

src/Plugin/Block/NodeHeaderBlock.php
```diff
@@ -0,0 +1,115 @@
...
+   * {@inheritdoc}
+   */
+  public function build() {
+    $build[] = [
+      '#theme' => 'page_header',
+      '#title' => $this->node->getTitle(),
+    ];
+
+    if (!$this->node->get('body')->isEmpty()) {
+      $body = $this->node->get('body')->first()->getValue();
```
Adnan-cds, line 88: This assumes that all nodes will have a "body" field.  My test site had a content type that doesn't have a body field and that led to a WSOD.  Worked fine otherwise.

config/optional/block.block.nodeheaderblock.yml
stephen-cox: This doesn't mean it will only work with the locagov theme, but it is needed for the localgov theme. If you were adapting the Croydon theme to work with this you could simple install the theme and place the block.

The dependency is there because this block is automatically enabled and placed when running and site install.

Adnan-cds: Hi Andy, unfortunately, that's how Drupal's block config files work.  You just have to mention a theme.  @see https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/modules/help_topics/config/optional/block.block.seven_help_search.yml @andybroomfield

Adnan-cds: Following the example of block.block_seven_help_search.yml, this file can be renamed to block.block.localgov_theme_nodeheaderblock.yml.  I cannot suggest anything other than that.

Adnan-cds: On a second thought, given this is an "optional" config file, the module info and composer.json files can skip requiring the localgov_theme.

andybroomfield: Ok, thanks. So it will be possible to place the block, even if localgov_theme is not installed.

finnlewis:
Happy to see that @Adnan-cds and @andybroomfield have looked at this. 

Code looks good, so merging to jolly things along!

Interested to know more about "the curse of lede" @Adnan-cds !

andybroomfield:
Confirming the tests work.

finnlewis:
Sorry @Adnan-cds @andybroomfield @stephen-cox - maybe I merged this too soon... I'm just aware that merging swiftly in the absence of major issues helps to keep things moving. 
Perhaps I should have actually tested the branch fully first. 
Happy to discuss and confirm what level of review we're all happy with.

andybroomfield:
@Adnan-cds do you want to raise this as an issue on this. I have written a test to verifiy the issue and can then have a look at a fix.

Adnan-cds:
> maybe I merged this too soon

Hi Finn, it's alright :) My closing message for the review wasn't very clear.  I should have clearly highlighted the bug in that message.  I will try to do better next time. @finnlewis 

> Happy to discuss and confirm what level of review we're all happy with.

Sure.  I tend to just point out things and leave it to the PR author to make the final call.  But happy to follow a pre-agreed approach.

PR #7, opened 2020-06-03
Fix WSOD with NodeHeaderBlock when on content type with no body field

### LINKED ISSUE #6: Bug: Node title block should work with nodes with no body field
We have managed to introduce a bug as part of #4 where nodes without a body field lead to WSOD.  @andybroomfield already has a fix, so here we go.

## How to reproduce
- Create a new content type in localgov site.
- Make sure the "body" field is present.
- Create a page of this new content type.
- View the page.
- This should lead to a WSOD.

andybroomfield:
Fix #6.

Checks that a node has a body field defined before setting a lede.
Adds test to verify.


stephen-cox:
Hi @andybroomfield Looks good to me. All tests pass. I found an error in the namespace for the test which was there when I created the file; which I've fixed.

PR #8, opened 2020-06-04
Change block Test not to rely on form submission.

ekes:
localgovdrupal/localgov_services#36 introduces a new compulsory field
to the localgov_services_page content type, this prevents this test
submitting the form.

Rather than track the form moved to using content creation trait,
with the minimum fields for this test.

### DIFF
--- a/tests/src/Functional/BlockTest.php
+++ b/tests/src/Functional/BlockTest.php
@@ -2,12 +2,18 @@
 
 namespace Drupal\Tests\localgov_core\Functional;
 
+use Drupal\node\NodeInterface;
 use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\node\Traits\ContentTypeCreationTrait;
+use Drupal\Tests\node\Traits\NodeCreationTrait;
 
 /**
  * Functional tests for LocalGovDrupal install profile.
  */
 class BlockTest extends BrowserTestBase {
+
+  use NodeCreationTrait;
+  use ContentTypeCreationTrait;
 
   /**
    * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


andybroomfield:
I prefer this method, though the test is no longer failing if the fix I applied for the missing body field is removed.

ekes:
Hmm... let's check that. As mentioned it was a rush re-write.

andybroomfield:
The fix is in `NodeHeaderBlock.php` to change
````php
if (!$this->node->get('body')->isEmpty()) {
````
to 
````php
if ($this->node->hasField('body') && !$this->node->get('body')->isEmpty()) {
````

ekes:
Ah it was because the content creation trait generates a body field -- a bit of refactoring too far. I'll put that bit back.

ekes:
@andybroomfield That should be correct now.

andybroomfield (approved):
Have checked and verified that the test fails when the body fix is missing, so these tests are checking the right thing.
Tests past with the fix.

andybroomfield:
I've been thinking that these may be a candidate to convert to kernal tests to imporve perfomance, as really its testing the block will send the right build array rather than the display on the page (which would be more of a profile test). Thats something for the future.

ekes:
Wrote up some suggestions triggered by this @stephen-cox @finnlewis @andybroomfield don't know what you think https://github.com/localgovdrupal/localgov/wiki/Testing suggest rather than starting an issue just hammer out details on the wiki?

ekes:
For the rest everyone good for a `squash and merge` for this one?

stephen-cox:
@ekes Thanks for you notes on testing; https://github.com/localgovdrupal/localgov/wiki/Writing-tests

I agree with the first point - we should only be testing things we want and if that doesn't include submitting a form then we should create entities directly. I think I had issues getting the syntax right to set specific field data, which is why I ended up submitting forms.

I'm don't think I agree with the second point. If we need to test form submission, but not the form widgets then posting an array of values seems a legitimate way of doing things - lots of the tests I looked at in core do this. Filling in each field in a form tests the widgets as well as form submission, which might not always be desirable.

PR #11, opened 2020-06-04
Media in localgov core

Adnan-cds:
## Media types and Media library

New submodule: localgov_core:localgov_media

New Media types:
- Document
- Image
- Remote video

Mostly lifted from the [demo_umami Drupal core profile](https://git.drupalcode.org/project/drupal/-/tree/8.9.x/core/profiles/demo_umami).

New image crop types:
- 3:2
- 7:3
- 8:7
- 16:9
- Square

Media images can be cropped using any of the above crop types.

Some **Image styles** have been modified to act on cropped images **where such a cropping has been applied**.  For example, when a Media image is cropped using the 3:2 cropping, the **Large 3:2** Image style will process that cropped image when asked to process the Media image.  But when no such cropping has been applied, *Large 3:2* will process the original Media image file.

~~I have not updated the **Responsive image styles** as they rely on breakpoints and the localgov_core module is not providing any breakpoint definition file at the moment.~~ The breakpoints now provided by localgov_media matches that of Bootstrap 4:
- Xtra small (Default)
- Small (576+px)
- Medium (768+px)
- Large (992+px)
- Xtra large (1200+px)

There are three Responsive image styles; again, copied from demo_umami:
- 3:2: Default image width is 768px; From Medium breakpoint onwards, width is 1536px
- Banner (21:9): Default image width is 768px; Medium and Large breakpoints use a width of 1440px; Extra large uses 2880px.
- Square: Default image width is 600px; From Medium breakpoint onwards, width of 900px is used.

Images will be [upscaled](https://en.wikipedia.org/wiki/Image_scaling) if necessary.  The *3:2* and *Square* Responsive image styles are used by the *Responsive 3x2* and *Square* **Display mode** of Media images.


Adnan-cds:
A quick way to manually test different Media image styles is to create a View of Media Images and then switch among different Entity **View modes**.  Once the image output is updated, copy the Image URL and inspect in a different browser tab or window.

finnlewis:
Following brief discussion with @ekes I removed the dependency on localgov_media from locagov_core.info.yml as we want to be able to support re-use of localgov_core and dependent modules without assuming they want our media presets.

finnlewis:
Installing works for me, just testing the phpunit tests. 
Also discussing with @ekes whether media entities need to be namespaced to localgov_

finnlewis:
Also, PHPUnit tests do pass, since merging in master to this branch.

ekes:
> Also discussing with @ekes whether media entities need to be namespaced to localgov_

Also crops and the like. They're common names that may exist in sites already.

But going to take the lead from Lightning here https://git.drupalcode.org/project/lightning_media They don't. You can opt in to each bit. But you opt-in, or you don't opt-in.

So this should stay very much optional. But if people opt in it takes over media.

finnlewis (approved):
I'm happy with this, it install OK and is optional. 
@ekes and I had a discussion about namespacing the media entities. 
He points out that if people in future have a site that they want to add this to, it would be safer to namespace the media entities like locagov_image localgov_document etc. 
I suggest that this is less likely to be an issue if it is optional, as they would not be running the install profile but using modules that have an optional dependency on this config. 
@ekes then also points out the Lightening don't namespace their entities.
https://git.drupalcode.org/project/lightning_media/-/tree/8.x-3.x/config/install

So let's go with this for now!

PR #18, opened 2020-06-16
Install and enable Fontawesome.

### LINKED ISSUE #13: Add Drupal Font Awesome Icons module
https://www.drupal.org/project/fontawesome

See discussion here: https://github.com/localgovdrupal/localgov_theme/issues/20

stephen-cox:
Discussion for this has been here: https://github.com/localgovdrupal/localgov_theme/issues/20

Closes #13

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,7 @@
     "license": "GPL-2.0-or-later",
     "minimum-stability": "dev",
     "require": {
+        "drupal/fontawesome": "^2.16",
         "drupal/image_widget_crop": "^2.3"
     }
 }
--- a/localgov_core.info.yml
+++ b/localgov_core.info.yml
@@ -8,3 +8,4 @@ dependencies:
   - drupal:block
   - drupal:field
   - drupal:node
+  - fontawesome:fontawesome


ekes:
Is this a core thing rather than a theme requirement? 8.9 introduces theme dependencies https://www.drupal.org/node/2937955

Adnan-cds:
We have paragraph fields and node fields that uses the FontAwesome module's "fontawesome_icon" field type.  So this is not entirely tied to the theme.

FAO @ekes

ekes:
> We have paragraph fields and node fields that uses the FontAwesome module's "fontawesome_icon" field type. So this is not entirely tied to the theme.

It would be nice to review that. It's quite a thing to require people to install if they are installing just one component.

finnlewis:
Interesting, so it seems the FontAwesome module is wanted by the theme and by one/some of the paragraphs, is that right?
If so, can we add it as a dependency to the theme and to the paragraphs module rather than core? 

Might there then be a further motivation to separate default paragraphs?

In fact, I can't see any reference to "fontawesome_icon" in the current https://github.com/localgovdrupal/localgov_paragraphs project, so maybe this is not needed yet but comes in on campaigns or homepage paragraphs? 

If so, @Adnan-cds is this something we can keep in the theme for now and bring in as a dependency to a future paragraphs module ?

ekes:
> If so, can we add it as a dependency to the theme and to the paragraphs module rather than core?
> Might there then be a further motivation to separate default paragraphs?

That was where my thoughts were going yesterday evening. It could be added as a dependency to the profile, but doesn't want to be added to core (as this gets installed with everything even if only one module is installed).

With the introduction of dependencies for themes though it makes even more sense to add the requirement where it is needed: the theme and the specific modules. No need to also put it on the profile as that will get the dependency in the chain; and no need for other sites to install it unless they install a module that needs it.

Breaking up paragraphs into submodules that can be enable or not depending also seems the way to go. It's expanding this outside of this issue, but the dependency of paragraphs on media had also already come up. If lots of modules will depend on _some_ but _not all_ shared paragraphs it would make sense to split off at least those paragraphs that add dependencies into submodules that can optionally be added. Media, and fontawesome thus far.

stephen-cox:
At the moment some of the services stuff uses Fontawesome icons, so arguably there's a dependency with core features, although not a very strong one.

My preference would be to create a Fontawesome library in core that can be included by modules that need it, but there appears to be a need with the upcoming homepage stuff to have this module. So if we don't include this module at the moment any code that includes a Fontawesome library would need to check if this module is installed or not.

At the moment we just want to fix the broken icons littering the site.

ekes:
Just checking an assumption I'm making here: If installing a module on another site you'd want the icons to fit whatever icon set you've already got on your site?

stephen-cox:
Not sure I fully follow what you're asking.

Fontawesome uses very specific classes and characters for it's icon set so I don't think you can just drop in another icon set without doing quite a bit of work.

Examples of services uses of Fontawesome:
- https://github.com/localgovdrupal/localgov_services/blob/master/modules/localgov_services_status/css/extra.css#L55
- https://github.com/localgovdrupal/localgov_services/blob/master/modules/localgov_services_landing/templates/node--localgov-services-landing--full.html.twig#L67

ekes:
Yes. That's theme specific output.

stephen-cox:
Further to our standup call, I'm going to close this can of worms and add a Fontawesome library to the theme. We can review what templates in modules should contain in due course and add the Fontawesome module when it's required.

PR #19, opened 2020-06-18
WYSIWYG text editor and format with Linkit.

### LINKED ISSUE #12: Add text format with CKeditor
At the moment there's only plaintext editing. People will want to use CKEditor.

What buttons do we want to include by default? 
Do we want to include an optional dependency to localgov_media with an entity browser for inserting media as part of this issue?

stephen-cox:
This largely follows what's done on the Croydon site. I have tweaked it a bit. We will want to confirm exactly what buttons the CKEditor should contain and write some tests, but I suggest we do that as a separate issue and get this deployed as Will likely want to use it when adding example content.

Closes  #12.

### DIFF
--- a/localgov_core.info.yml
+++ b/localgov_core.info.yml
@@ -6,5 +6,7 @@ core_version_requirement: ^8.8 || ^9
 
 dependencies:
   - drupal:block
+  - drupal:editor
   - drupal:field
   - drupal:node
+  - linkit:linkit
--- a/localgov_core.info.yml
+++ b/localgov_core.info.yml
@@ -6,7 +6,5 @@ core_version_requirement: ^8.8 || ^9
 
 dependencies:
   - drupal:block
-  - drupal:editor
   - drupal:field
   - drupal:node
-  - linkit:linkit
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,7 @@
     "license": "GPL-2.0-or-later",
     "minimum-stability": "dev",
     "require": {
-        "drupal/image_widget_crop": "^2.3"
+        "drupal/image_widget_crop": "^2.3",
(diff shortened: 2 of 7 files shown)


localgov_core.info.yml
```diff
@@ -6,5 +6,7 @@ core_version_requirement: ^8.8 || ^9
 
 dependencies:
   - drupal:block
+  - drupal:editor
```
ekes, line 9: Should these be in `localgov_media.info.yml`

ekes:
Makes sense now. 

Also makes the case for possibly spinning off the media module, but can be looked at as thing develop.

stephen-cox:
Checks are passing. There's an issue with running tests in GitHub Actions that sometimes throws the following error for the first test or two. Will need investigating separately.
```
Drupal\Tests\localgov\Functional\LocalGovProfile
 ✘ Local gov drupal profile
   │
   │ Undefined index: value
   │ 
   │ /var/www/html/web/core/includes/install.core.inc:2296
   │ /var/www/html/web/core/includes/install.core.inc:1082
   │ /var/www/html/web/core/includes/install.core.inc:695
   │ /var/www/html/web/core/includes/install.core.inc:570
   │ /var/www/html/web/core/includes/install.core.inc:118
   │ /var/www/html/web/core/lib/Drupal/Core/Test/FunctionalTestSetupTrait.php:296
   │ /var/www/html/web/core/tests/Drupal/Tests/BrowserTestBase.php:573
   │ /var/www/html/web/core/tests/Drupal/Tests/BrowserTestBase.php:406
```

finnlewis:
@stephen-cox this looks good to me. I tested it locally, and linkit works within the CKEditor.
Just re-ran the tests... and all have passed!

PR #23, opened 2020-07-01
Add and patch Field Group module

### LINKED ISSUE #22: Add Field Group module as a dependency of core and patch for d.o. issue #3059614
It's currently beings installed through composer by the a few LocalGiv modules which uses slightly different version numbers. It would be better to include it once in core and remove it from the other modules - which depend on core anyway.

- localgov_alert_banner
- localgov_paragraphs
- localgov_services

There's also an issue with Field Group which means it gives a warning on node edit pages, which cause tests that visit this page to fail: https://www.drupal.org/project/field_group/issues/3059614

stephen-cox:
Closes #22

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,6 @@
     "license": "GPL-2.0-or-later",
     "minimum-stability": "dev",
     "require": {
-        "cweagans/composer-patches": "^1.6",
         "drupal/field_group": "^3.1",
         "drupal/image_widget_crop": "^2.3",
         "drupal/linkit": "^6.0-beta1"
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "minimum-stability": "dev",
     "require": {
         "cweagans/composer-patches": "^1.6",
-        "drupal/field_group": "~3.0",
+        "drupal/field_group": "^3.1",
         "drupal/image_widget_crop": "^2.3",
         "drupal/linkit": "^6.0-beta1"
     },
@@ -18,7 +18,7 @@
             "drupal/core": "-p2"
         },
         "patches": {
-            "drupal/geolocation": {
+            "drupal/field_group": {
                 "Fix 'Undefined property: stdClass::$region' error #3059614": "https://www.drupal.org/files/issues/2020-06-15/3059614-37.patch"
             }
         }
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,21 @@
     "license": "GPL-2.0-or-later",
     "minimum-stability": "dev",
     "require": {
+        "cweagans/composer-patches": "^1.6",
+        "drupal/field_group": "~3.0",
         "drupal/image_widget_crop": "^2.3",
         "drupal/linkit": "^6.0-beta1"
+    },
+    "extra": {
+        "enable-patching": true,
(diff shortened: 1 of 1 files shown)


ekes:
Only reservation here is what happens if field_group is already installed?

Also if we've tested this should add a follow-up review on d.o and `reviewed and tested by the community`. It might make that complexity go away already :)

stephen-cox:
What's the problem with Field Group being installed already?

You can test this by checking out the feature/22-field-group-module branch of the localgov_project module. This in turn uses the feature/22-field-group-module branch of the profile that uses a number of branches of different modules to resolve circular dependency hell.

ekes:
It was a question about what happens if field group is installed already. I've now done it on stock drupal + field_group. It doesn't seem to be a problem but it also doesn't seem to patch:-

```
➜  test composer require localgovdrupal/localgov_core:dev-feature/22-field-group-module 
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 5 installs, 0 updates, 0 removals
  - Installing cweagans/composer-patches (1.6.7): Downloading (100%)         
  - Installing drupal/linkit (6.0.0-beta1): Loading from cache
  - Installing drupal/crop (2.1.0): Loading from cache
  - Installing drupal/image_widget_crop (2.3.0): Loading from cache
  - Installing localgovdrupal/localgov_core (dev-feature/22-field-group-module 098f213): Cloning 098f213ddc from cache
Writing lock file
Generating autoload files
28 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
```

stephen-cox:
This is only needed to suppress a warning message and allow certain tests to pass. There's a possibility that not patching existing installs may mean that those tests will fail. Given there are a no installations in production, forcing a re-install of the module (by deleting it) so the patch applies shouldn't be too much of an issue.

Getting the patch to apply was an absolute ballache due to caching; locally in Composer and also at Packagist. My understanding is that `cweagans/composer-patches` needs to be included at the top level and subsequent levels for patching to work. That's what we're doing here: https://github.com/localgovdrupal/localgov_paragraphs/blob/15b1e554f5fc4dd7b68dce266ddbc3db6049f44e/composer.json#L9

composer.json
```diff
@@ -6,7 +6,6 @@
     "license": "GPL-2.0-or-later",
     "minimum-stability": "dev",
     "require": {
-        "cweagans/composer-patches": "^1.6",
```
ekes: So this line isn't needed. 

If it's been included an enabled in a parent composer.json it will work (see caveat about install order in comment coming).
If the parent composer.json doesn't have it, requiring it doesn't make sense.

ekes:
The caveat mentioned above is:-

If field_group has already been required and installed it won't patch when you require localgov_core.
It will however patch it when you _then_ require something else, as it seems to recompute the patches then with the localgov_core one included!

And repeating the comment with the code above if field_group hasn't been already installed it will patch it on the way in when you require localgov_core (if you have the requirements for patching in the parent composer.json).

stephen-cox:
Thanks @ekes This is all working for me and needs to be merged for the campaigns stuff, so merging and we can review later on if necessary.

PR #24, opened 2020-07-03
Add default workflow, assigned to roles and permissions. Optional.

### LINKED ISSUE #42: Shared: workflow
There are just two workflows: editorial and news.

Assume there will be default workflows in a localgov_workflow module? @finnlewis 

I also assume the workflows for lots of content types are much the same? @bhccwebmaster @willcallaghan Looking at croydon lots of content types have workflow and they all seem to be 'editorial'.

First dependencies here are services sublanding and services page (not landing page interestingly). I know there was some discovery to do about where workflows weren't working @andybroomfield ? 

But to begin with the two that exist into a standalone module?

ekes:
Closes localgovdrupal/localgov#42

 * Add the editorial workflow to localgov_ content types if they don't have
   any other workflow.
 * Add the default roles.
 * Assign transition permissions to those roles.

All optional, no dependencies, all can be overriden.


ekes:
I've got cross-eyed by this point, so check carefully :-)

modules/localgov_workflows/localgov_workflows.info.yml
```diff
@@ -0,0 +1,9 @@
+name: LocalGovDrupal Workflows
+description: Provides the default editorial workflow for content types.
+package: LocalGovDrupal
+type: module
+core_version_requirement: ^8.8 || ^9
+
+dependencies:
+  - drupal:content_moderation
+  - localgovdrupal:localgov_core
```
stephen-cox, line 9: Just a question more than anything else, what should the namespace of a dependency be? This works fine, but I would have used `main_module:sub_module` so `localgov_core:localgov_core`. Whatever we decide we should be consistent in it's usage.

stephen-cox (changes requested):
Installing and uninstalling the workflow module adds and removes the workflow and it seems to work fine as user 1.

The roles module is not a dependency of workflow and needs to be enabled separately, but this doesn't result in permissions for the roles being set if roles is enabled after workflow, so it would seem the roles module is required.

I think 'Comms designer' should be content designer.

Neither of the roles have enough permissions to create content. Does this need to be considered here or as part of a wider review of roles and permissions?

The actual workflow should be approved by content people, but that can be done after this pull request is approved.

modules/localgov_workflows/localgov_workflows.info.yml
ekes: Think I was just getting confused between composer dependencies and drupal dependencies. It is `main_module:sub_module` so `localgov_core:localgov_core`

ekes:
> I think 'Comms designer' should be content designer.

Should actually be 'Content designer' as I think was the final edit
https://github.com/localgovdrupal/localgov/issues/42#issuecomment-648847397
So that can be updated.

I left the machine name as localgov_editor as it's expected that the user names can be changed anyway.

> Neither of the roles have enough permissions to create content. Does this need to be considered here or as part of a wider review of roles and permissions?

So far I just included what would be included in a self contained workflow module - they could even be in a role.yml like this. But my expectation was that these might be roles (optionally) shared by other modules. So yes I think the default roles/permissions need to be thought out. The absolute simplest would be to keep roles within the their module domains, and then users would have to have several: services creator, workflow editor, etc. but I don't know that's desirable.

stephen-cox:
Just thinking about how to test this - do we want to add a basic bunch of permissions to allow editors to create and edit service nodes to the roles module? These can be reviewed, but will make testing stuff tomorrow easier.

finnlewis:
We've not looked at this for a while. 
Passing to @ekes to find a way forward.

stephen-cox:
@ekes My suggestion is to add a few content editing permissions for service pages to the localgov_roles module. These can then be reviewed in due course. I can get this done for review tomorrow if you think this is okay for now?

stephen-cox (approved):
@ekes I've added some permissions to the localgov_roles module so that it's now possible to test the workflow out of the box.

PR #25, opened 2020-07-10
Page header title block

stephen-cox:
This pull request defines a PageHeaderBlock that can be used for displaying page titles and ledes across the site. It also provides a PageHeaderDisplayEvent that modules can subscribe to and use to change the title and lede and to hide the block.

Examples of subscribing to the PageHeaderDisplayEvent and modifying the display of this block can be seen:

- Change title and lede on guide pages: https://github.com/localgovdrupal/localgov_guides/pull/21
- Hide block on campaign overviews and campaign pages:  https://github.com/localgovdrupal/localgov_campaigns/pull/27

Discussion of why this block exists and does what is does is here: https://github.com/localgovdrupal/localgov/issues/87

This requires tests before merging so marking as WIP.

I'm concerned that this may be over engineering the problem and that there maybe easier ways to do this, so creating this pull request for a discussion. All thoughts and criticism welcome.

To test this and the 2 related pull request checkout the [feature/localgov-87-title-block](https://github.com/localgovdrupal/localgov_project/tree/feature/localgov-87-title-block) in the project directory, run a composer update and a Drupal site install. Something like:
```bash
git checkout feature/localgov-87-title-block
composer update --no-cache
lando drupal si localgov
lando drupal en localgov_demo
```


src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -0,0 +1,230 @@
...
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    $this->currentRouteMatch = $current_route_match;
+    $this->eventDispatcher = $event_dispatcher;
+    $this->requestStack = $request_stack;
+    $this->titleResolver = $title_resolver;
+
+    // Find the entity, if any, associated with the current route.
+    // Adapted from https://www.computerminds.co.uk/drupal-code/get-entity-route
+    if (($route = $this->currentRouteMatch->getRouteObject()) && ($parameters = $route->getOption('parameters'))) {
```
ekes, line 114: I need to think about this more generally, and look at the title resolver service, and what the options might be for template preprocess or something. I'm sure you've thought them through. Just a double think. But anyway... while reading this code. I was thinking.

Why getting the raw parameters and checking for the keys and so forth?

foreach ($route_match->getParameters() as $paramater) { if ($parameter instanceof EntityInterface && ...)

Also why does it need a canonical link template?

Sorry just getting my head round it.

stephen-cox: So this is the danger of just copying and pasting code (although I did include a link to where I found it).

Digging into the Drupal source code; this is largely copied from the content translation module: https://git.drupalcode.org/project/drupal/-/blob/8.9.x/core/modules/content_translation/content_translation.module#L662. 

Git blame points to the main logic for this here: https://www.drupal.org/project/drupal/issues/2303525#comment-9032709 - which doesn't give any reasons as to why it's done in the way it is.

We can certainly remove the canonical link template as that's only needed for translated content: https://www.drupal.org/project/drupal/issues/2479377

src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -0,0 +1,230 @@
...
+      return Cache::mergeTags(parent::getCacheTags(), $this->entity->getCacheTags());
+    }
+    return parent::getCacheTags();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    return Cache::mergeContexts(parent::getCacheContexts(), ['route']);
```
Adnan-cds, line 227: +1 for switching from **url.path** to **route**.  We have run into a bizarre edge case recently that was caused by url.path.  It was not really a bug, more of a peculiarity of the hosting environment.  But **route** survived fine where url.path didn't.

src/Plugin/Block/PageHeaderBlock.php
stephen-cox: > Why getting the raw parameters and checking for the keys and so forth?
> 
> foreach ($route_match->getParameters() as $paramater) { if ($parameter instanceof EntityInterface && ...)

This seems the easiest way of determining the whether the route references an entity. `$this->currentRouteMatch->getParameters()` is also an option, but it returns more parameters and checking whether any of them is an entity looks to be more involved.

> Also why does it need a canonical link template?

It doesn't. Removed

stephen-cox:
Would be nice to get this merged for review tomorrow. I will do so tomorrow morning unless there are any objections raised before then.

PR #27, opened 2020-08-03
Added home page welcome block

danchamp:
This PR adds a block for a home page message to be displayed on installation of LocalGov Drupal. Provides default content.

Used by install profile as part of simple home page, see [/localgovdrupal/localgov/issues/96](/localgovdrupal/localgov/issues/96)

### DIFF
--- a/src/Plugin/Block/HomeWelcomeBlock.php
+++ b/src/Plugin/Block/HomeWelcomeBlock.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Drupal\localgov_core\Plugin\Block;
+
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Block\BlockPluginInterface;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides a 'Powered by LocalGovDrupal' block.
+ *
+ * @Block(
+ *   id = "localgov_home_welcome_block",
+ *   admin_label = @Translation("Home welcome block")
+ * )
+ */
+class HomeWelcomeBlock extends BlockBase implements BlockPluginInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return ['label_display' => FALSE];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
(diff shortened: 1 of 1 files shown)


src/Plugin/Block/HomeWelcomeBlock.php
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!empty($config['localgov_home_welcome_message'])) {
+      $message = $config['localgov_home_welcome_message'];
+    }
+    else {
+      $message = $this->t('<h2>You have just installed a LocalGov Drupal site</h2>');
+    }
+
+    return [
+      '#markup' => $message['value'],
```
Adnan-cds, line 41: This is throwing the following PHP exception following a fresh install: "Error: Cannot use object of type Drupal\Core\StringTranslation\TranslatableMarkup as array in Drupal\localgov_core\Plugin\Block\HomeWelcomeBlock->build() (line 41 of modules/contrib/localgov_core/src/Plugin/Block/HomeWelcomeBlock.php)." :(  Replacing "$message['value']" with "$message" fixes this.

danchamp: Thanks Adnan, that exception is fixed now.

> Because this block is tied to the localgov install profile, perhaps that would be a better home for this block rather than localgov_core module.

Ah, I wasn't aware install profiles could include anything other than config and translations - it did feel odd to have to put this in localgov_core. I'll move it over to the install profile and delete this branch, that makes much more sense.

Adnan-cds: Thanks Dan.  Just noticed that @finnlewis has also written "This can sit in the locagov install profile." in [Simple home page - place holder](https://github.com/localgovdrupal/localgov/issues/96).  So there is now a stronger case for moving this code to the localgov install profile.

danchamp:
Closing PR and moving block class to LocalGov install profile as per @Adnan-cds [comment](/localgovdrupal/localgov_core/pull/27#pullrequestreview-459959444)

See install profile PR: [/localgovdrupal/localgov/pull/98](/localgovdrupal/localgov/pull/98)

PR #33, opened 2020-09-01
Media view modes for the Homepage Paragraphs

Adnan-cds:
Hi Steve,
I have added three Entity view modes for the Media image: Newsroom teaser, Featured, Featured large.  The News image linked from the Newsroom teaser Paragraph can be displayed using these new view modes.

Task: https://github.com/localgovdrupal/localgov/issues/107

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,9 @@
     "require": {
         "drupal/field_group": "^3.1",
         "drupal/image_widget_crop": "^2.3",
-        "drupal/linkit": "^6.0-beta1"
+        "drupal/linkit": "^6.0-beta1",
+        "drupal/pathauto": "^1.8",
+        "drupal/token": "^1.7"
     },
     "extra": {
         "enable-patching": true,
--- a/modules/localgov_media/config/optional/core.entity_view_display.media.image.localgov_featured.yml
+++ b/modules/localgov_media/config/optional/core.entity_view_display.media.image.localgov_featured.yml
@@ -1,0 +1,29 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.media.localgov_featured
+    - field.field.media.image.field_media_image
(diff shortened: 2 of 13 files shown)


finnlewis:
@stephen-cox @Adnan-cds sorry, I'm a bit out of the loop, is there any debate about whether these live here or is this all good?

Adnan-cds:
Hello Finn,
In the Brighton codebase, the "Newsroom teaser" Paragraph has three display modes who all do the same thing.  The only difference is which image styles they use on the uploaded image.  In LocalGov, we are using Media images instead of uploading images directly.  When porting the Newsroom teaser Paragraph, I have used these three Media image display modes on a Media image field instead of the three original image styles on an image field.

The "Newsroom teaser" Paragraph is used to build the Homepage which is at the moment housed in the localgov_homepage_paragraphs module.  Initially I kept these display modes there, but they looked useful beyond the "Newsroom teaser" Paragraphs.  So I have raised this pull request to move these to the localgov_media module.

The [pull request for the homepage paragraphs](https://github.com/localgovdrupal/localgov_paragraphs/pull/6) depends on this pull request.  You can check the [discussion on the homepage](https://github.com/localgovdrupal/localgov_paragraphs/pull/6) although that's not entirely relevant to this pull request.

FAO @finnlewis @stephen-cox

stephen-cox (approved):
The only issue I can see with this is that it further cements the dependency between anything that relies on localgov_paragraphs and localgov_media.

There was an idea that it might be possible to use some modules (such as localgov_campaigns) with your own media settings, but the reality is that something like Campaigns if so media rich that making all the media stuff optional is going to be a pain.

Happy to approve this. We can unpick the media dependencies between paragraphs and other LocalGov modules later if it required.

Adnan-cds:
Thanks for reviewing this Steve.  I am merging this now :)

FAO @stephen-cox

PR #42, opened 2020-09-30
Add tables button to CKEditor config for default WYSIWYG text format.

finnlewis:
Does what it says on the tin!

### DIFF
--- a/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
@@ -48,12 +48,16 @@ settings:
           name: Tools
           items:
             - Source
+        -
+          name: Tables
+          items:
+            - Table
   plugins:
+    language:
+      language_list: un
     drupallink:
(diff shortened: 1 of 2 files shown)


Adnan-cds (approved):
Works fine for me :)  We inherited the CKEditor Table button from BHCC, so I am assuming Andy is fine with this change too.

In the future we may want to create a **button group** for this new Table button and any other buttons inserting complex markup.

@andybroomfield @finnlewis

andybroomfield:
We actually have specific markup for our tables, assume theme will auto make tables responsive with bootstrap or we can reinsert the required markup ourselves.

(From memory the table element needs to add a .table class, which is then picked up to drop the table into a list on narrow viewports).

Apologies been a bit quiet after coming back from my leave, currently BHCC is in the process of switching off old website and having to deal with broken issues.

andybroomfield:
Example page with responsive table on BHCC

https://new.brighton-hove.gov.uk/allotments/get-allotment/allotment-plot-types

PR #52, opened 2021-03-29
remove hardcoded classes from page title template that are not relevant to all themes

markconroy:
This PR
1. Removes hard-coded classes from the `classes` array. These classes pre-suppose a specific CSS structure (I'm guessing Bootstrap), such as "clear-both".
2. Removes hard-coded `.header` class from the page title. This class should only be used for the main page header, so we can style that via a class instead of an element. Currently, having this class in this template is causing all my page titles to inherit the same CSS that my site header has!
3. Adds in generic classes that can be used by any theme, using BEM naming standards, based on a BEM block called `page-title` (I hope this doesn't conflict with the `.page-title` class in the core `page-title.html.twig` file, but I am guessing this template overrides that one.

I think it would probably be better to clone and manipulate the `page-title.html.twig` template rather than this very specific replacement. But we can discuss that on the technical group chats.

### DIFF
--- a/templates/localgov-page-header-block.html.twig
+++ b/templates/localgov-page-header-block.html.twig
@@ -8,16 +8,18 @@
 * @ingroup themeable
 */
 #}
-{% set classes = [
-    'clear-both',
-    'headers',
-    'key',
-] %}
+
+{% 
+  set classes = [
+    'page-title',
+  ] 
+%}
+
 {% if title %}
-<div{{ attributes.addClass(classes) }}>
-  <h1 class="header">{{ title }}</h1>
-  {% if lede %}
-  <div class="subheader">{{ lede }}</div>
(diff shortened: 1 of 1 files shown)


markconroy:
Closing and re-opening PR to force tests to run again.

finnlewis:
@markconroy just a note to check if this still makes sense.

markconroy:
I've made some minor changes to the template now - it now has the same markup as we have in `localgov_base` theme, which I think makes sense.

finnlewis:
@markconroy to take another look?

finnlewis:
Hiya @markconroy - is this pull request still valid or shall we close it off?

markconroy:
Let's close it, the PR comes from an old repo. We can recreate from the core repo later if needed.

PR #54, opened 2021-04-02
allow <br> in wysiwyg

### LINKED ISSUE #53: Allow <br> in WYSIWYG input
Currently if we add a line break (shift+enter), the `<br>` is in the source, but doesn't get converted to a `<br>` when output.

Let's switch this on in the WYSIWYG input settings.

markconroy:
Closes #53

### DIFF
--- a/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
@@ -55,11 +55,11 @@ settings:
   plugins:
     language:
       language_list: un
+    stylescombo:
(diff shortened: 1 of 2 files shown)


ekes:
How's this in media. Anyway...

I guess this good, unless it's a problem if sites were to get the update and they have <br> in their markup that aren't presently being displayed?

markconroy:
If someone has a br in their markup and it's not being displayed and they think that's correct, then that probably shouldn't be a br. In that case, they need to fix their markup.

If someone has a br in their markup and it's not being displayed and they think that is incorrect, then we need to fix how we treat br. In that case, this PR is correct.

Without this br, there is no way to create line breaks in our wysiwyg.

ekes:
Yer I realized after I went away from my computer this is also the install config.

PR #65, opened 2021-04-29
*drush deploy* compatible field renaming

Adnan-cds:
At the moment, field renaming involves exporting configuration files in the *live* site after running the relevant database updates.  This is different from the usual deployment process where database updates are run in a development machine, configs are exported and then *drush deploy* or equivalent is used by a CI/CD system to deploy Drupal in the live site.

The current process is necessary because every time the database updates for field renaming are run, they generate **different UUID** values for field configs which instantly makes the new field config out-of-sync with any config file on the disk.  To avoid this, we should source the UUIDs from config files on the disk *when* such files are present.  This makes the process *drush deploy* compatible.

### How to test
- drush --yes updb
- drush --yes config:export
- drush --yes sql:drop; drush sql:query --file=/path/to/database/dump.sql
- drush deploy

This, as expected, would wipe out any data from all the new fields (e.g. localgov_twitter, localgov_page_components, etc.) **unless** this minor change has been applied.  This change avoids the data loose scenario as it sources the UUIDs from config files on the disk wherever possible.

### DIFF
--- a/src/FieldRenameHelper.php
+++ b/src/FieldRenameHelper.php
@@ -51,6 +51,12 @@ class FieldRenameHelper {
     unset($new_field_storage['_core']);
     $new_field_storage['field_name'] = $new_field_name;
     $new_field_storage['id'] = str_replace($old_field_name, $new_field_name, $new_field_storage['id']);
+    // Useful where config file for this new field storage has been already
+    // exported after database update in a *development* site.
+    if ($existing_field_storage_config = \Drupal::service('config.storage.sync')->read("field.storage.{$new_field_storage['id']}")) {
+      $new_field_storage['uuid'] = $existing_field_storage_config['uuid'] ?? '';
+    }
+
     $new_field_storage = FieldStorageConfig::create($new_field_storage);
     $new_field_storage->original = $new_field_storage;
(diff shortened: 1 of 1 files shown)


ekes:
@andybroomfield I've no picture of how the field naming worked. I could dive in, but does the check make sense to you?

andybroomfield (approved):
I tested these and they don't break the existing updates which is good. 
I was a little confused why we are getting the new fields UUID and not trying to reuse the old one. It wasn't till I kept a preserved version of the config exported, did a fresh install and rename that I saw the UUIDs from the config I exported was preserved.
... So if I understand correctly, if you had got the config files in config/sync (eg. from a git push) then the fields get renamed and the uuid from the config file gets used. If not a new (eg. your on your local / development environment) the UUID is generated, and if you happend to have a field with the same name already in the DB it fails gracefully notifing you that field cannot be renamed.

Adnan-cds:
> if you had got the config files in config/sync (eg. from a git push) then the fields get renamed and the uuid from the config file gets used. If not a new (eg. your on your local / development environment) the UUID is generated, 

That's right.  Sorry, I couldn't have been clearer in my original description.

Thanks for the approval.  Merging...

PR #68, opened 2021-05-05
Drupal 9 coding standards and deprecation fixes

### LINKED ISSUE #67: Drupal 9 coding standards and deprecation issues
## Coding standards

```
FILE: ...e_header_event_test/src/EventSubscriber/PageHeaderSubscriber.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 13 | WARNING | The class short comment should describe what the
    |         | class does and not simply repeat the class name
----------------------------------------------------------------------
```

## Deprecated code

```
------ ---------------------------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_core/modules/localgov_workflows/tests/src/Kernel/WorkflowsInstallTest.php  
 ------ ---------------------------------------------------------------------------------------------------- 
  78     Call to deprecated method assertEqual() of class Drupal\KernelTests\KernelTestBase:                 
         in drupal:8.0.0 and is removed from drupal:10.0.0. Use                                              
           $this->assertEquals() instead.                                                                    
  79     Call to deprecated method assertEqual() of class Drupal\KernelTests\KernelTestBase:                 
         in drupal:8.0.0 and is removed from drupal:10.0.0. Use                                              
           $this->assertEquals() instead.                                                                    
  88     Call to deprecated method assertEqual() of class Drupal\KernelTests\KernelTestBase:                 
         in drupal:8.0.0 and is removed from drupal:10.0.0. Use                                              
           $this->assertEquals() instead.                                                                    
  89     Call to deprecated method assertEqual() of class Drupal\KernelTests\KernelTestBase:                 
         in drupal:8.0.0 and is removed from drupal:10.0.0. Use                                              
           $this->assertEquals() instead.                                                                    
  90     Call to deprecated method assertEqual() of class Drupal\KernelTests\KernelTestBase:                 
         in drupal:8.0.0 and is removed from drupal:10.0.0. Use                                              
           $this->assertEquals() instead.                                                                    
 ------ ---------------------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_core/src/Event/PageHeaderDisplayEvent.php                                                         
 ------ --------------------------------------------------------------------------------------------------------------------------- 
  10     Class Drupal\localgov_core\Event\PageHeaderDisplayEvent extends deprecated class Symfony\Component\EventDispatcher\Event:  
         since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead                                                   
 ------ --------------------------------------------------------------------------------------------------------------------------- 
```

stephen-cox:
Closes #67 

Part of https://github.com/localgovdrupal/localgov/issues/197

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,7 +52,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: localgovdrupal/drupal-container
-          ref: master
+          ref: php${{ matrix.php-version }}
 
       - name: Create LocalGov Drupal project
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: Test LocalGov Drupal Core
 
 on:
   push:
-    branches: [ master ]
+    branches: [ 2.x ]
   pull_request:
-    branches: [ master ]
+    branches: [ 2.x ]
 
 jobs:
 
@@ -32,7 +32,7 @@ jobs:
           ref: master
 
       - name: Create LocalGov Drupal project
-        run: composer create-project --stability dev localgovdrupal/localgov-project ./html
+        run: composer create-project --stability dev localgovdrupal/localgov-project ./html "2.x"
 
       - name: Save git branch and git repo names to env if this is not a pull request
         if: github.event_name != 'pull_request'
@@ -52,7 +52,7 @@ jobs:
           composer global config github-oauth.github.com ${{ github.token }}
 
       - name: Obtain the test target from the repo that triggered this workflow
-        run: composer --working-dir=html require --with-all-dependencies localgovdrupal/${{ github.event.repository.name }}:"dev-${GIT_BRANCH} as 1.0.x-dev"
+        run: composer --working-dir=html require --with-all-dependencies localgovdrupal/${{ github.event.repository.name }}:"dev-${GIT_BRANCH} as 2.x-dev"
 
   phpcs:
     name: Coding standards checks
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,11 @@ name: Test LocalGov Drupal Core
 
 on:
   push:
-    branches: [ 2.x ]
+    branches:
+      - '2.x'
   pull_request:
-    branches: [ 2.x ]
+    branches:
+      - '2.x'
 
 jobs:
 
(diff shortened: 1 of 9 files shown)


stephen-cox:
Hmm ... chicken and egg problem.

Drupal 9 code deprecation
```
Class Drupal\localgov_core\Event\PageHeaderDisplayEvent extends deprecated class Symfony\Component\EventDispatcher\Event:  since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead 
```

But in Drupal 8 tests fail with
```
Class Symfony\Contracts\EventDispatcher\Event not found.
```

andybroomfield:
Not sure its relevant here, but when debugging the the scheduler module, I came across this code.
````php
/**
   * Dispatch a Scheduler event.
   *
   * All Scheduler events should be dispatched through this common function.
   *
   * Drupal 8.8 and 8.9 use Symfony 3.4 and from Drupal 9.0 the Symfony version
   * is 4.4. Starting with Symfony 4.3 the signature of the event dispatcher
   * function has the parameters swapped round, the event object is first,
   * followed by the event name string. At 9.0 the existing signature has to be
   * used but from 9.1 the parameters must be switched.
   *
   * @param \Drupal\Component\EventDispatcher\Event $event
   *   The event object.
   * @param string $event_name
   *   The text name for the event.
   *
   * @see https://www.drupal.org/project/scheduler/issues/3166688
   */
  public function dispatch(Event $event, string $event_name) {
    // \Symfony\Component\HttpKernel\Kernel::VERSION will give the symfony
    // version. However, testing this does not give the required outcome, we
    // need to test the Drupal core version.
    // @todo Remove the check when Core 9.1 is the lowest supported version.
    if (version_compare(\Drupal::VERSION, '9.1', '>=')) {
      // The new way, with $event first.
      $this->eventDispatcher->dispatch($event, $event_name);
    }
    else {
      // Replicate the existing dispatch signature.
      $this->eventDispatcher->dispatch($event_name, $event);
    }
  }
````

andybroomfield:
The order of event dispatcher parameters is different between Symfony 3 and Symfony 4.4, which Drupal core 9.1 uses. So this becomes one of the issues where we decide if we need to have explicit support for both Drupal 8 and 9 in this module.

I'm coming round to having explicit releases that are Drupal 9 only now (in a v2.0 branch), depending on if people feel this would simplify the maintenance, or more complicate things with maintaining both releases (I'd like to see the Drupal 8 v1.0 versions at least maintained and available till Drupal 8 end of life in November).

ekes:
This won't yet pass, as we need to sort the project and profile - and which project gets deployed by the test runner; but hopefully should be able to test it against just Drupal 9 in a bit...

ekes:
Should we get these into 2.x now, so we can then run the tests against them?

finnlewis (approved):
Looks good to me (and we agreed that we could review internally for this round of pull requests)

PR #76, opened 2021-05-20
Update paragraph tables with references to new field names

### LINKED ISSUE #75: Field renaming is wiping out Paragraphs
Hello,
We have run the [field renaming update](https://github.com/localgovdrupal/localgov_theme/releases/1.0.0-alpha4) yesterday.  Fields were renamed correctly without any issues.  Unfortunately, some Paragraph entities referenced by the field_topics field (now localgov_topics) became orphaned in the process.  These were later purged by the [Entity reference revisions module](https://git.drupalcode.org/project/entity_reference_revisions/-/blob/8.x-1.x/entity_reference_revisions.module#L310).  Because it uses a [queue worker](https://git.drupalcode.org/project/entity_reference_revisions/-/blob/8.x-1.x/src/Plugin/QueueWorker/OrphanPurger.php), it wasn't until this morning that the wiping was executed.  This left me puzzled as to the cause given that all looked fine immediately after deployment.  Eventually I had to restore some lost rows of the relevant tables from an older database backup to bring back normalcy.

The effected Paragraph was [topic_list_builder](https://github.com/localgovdrupal/localgov_services/blob/2.x/modules/localgov_services_sublanding/config/install/paragraphs.paragraphs_type.topic_list_builder.yml) bundled with the localgov_service_sublanding page submodule.  But it can happen to any Paragraph referenced from a renamed Entity reference revision field.

I don't see any quick solution other than not to [rename field_topics](https://github.com/localgovdrupal/localgov_services/blob/6ee3adc340647f647f56ee445bfae5e1d57018fa/modules/localgov_services_sublanding/localgov_services_sublanding.install#L23).

@andybroomfield @finnlewis @tom-steel

andybroomfield:
Fix #75

If the field storage is a paragraph, an extra step is executed:
  The tables `paragraphs_item_field_data` and
  `paragraphs_item_revision_field_data` are updated for the entity type
  and source field name to the new field name.

eg. updated parent_type = 'node' and parent_field_name = 'field_topics'
is updated to parent_field_name = 'localgov_topics'

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,9 @@
         "drupal/redirect": "^1.6",
         "drupal/token": "^1.7"
     },
+    "require-dev": {
+        "drupal/paragraphs": "^1.12"
+    },
     "extra": {
         "enable-patching": true,
         "composer-exit-on-patch-failure": true,
--- a/localgov_core.info.yml
+++ b/localgov_core.info.yml
@@ -9,3 +9,6 @@ dependencies:
   - drupal:field
   - drupal:node
   - field_group:field_group
+
+test_dependencies:
+  - paragraphs:paragraphs
--- a/src/FieldRenameHelper.php
+++ b/src/FieldRenameHelper.php
@@ -7,6 +7,7 @@ use Drupal\field\Entity\FieldStorageConfig;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\Core\Entity\Entity\EntityViewDisplay;
 use Drupal\Core\Entity\Entity\EntityFormDisplay;
+use Drupal\Core\Utility\UpdateException;
 
 /**
  * Renames an existing entity field.
@@ -45,6 +46,14 @@ class FieldRenameHelper {
       return;
     }
 
+    // Exception if trying to update an entity reference revisions field that
+    // is not a paragraph.
+    $field_type = $field_storage->get('type');
+    $field_target_type = $field_storage->get('settings')['target_type'] ?? NULL;
+    if ($field_type == 'entity_reference_revisions' && $field_target_type != 'paragraph') {
+      throw new UpdateException('Entity referenece revisions fields except for paragraphs cannot be renamed.');
+    }
+
     // Create new field storage.
     $new_field_storage = $field_storage->toArray();
     unset($new_field_storage['uuid']);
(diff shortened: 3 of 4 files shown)


andybroomfield:
Will also need to be ported to 2.x branch?

andybroomfield:
First pass @Adnan-cds Let me know if this works for you.

Adnan-cds:
Hi Andy,
The parent field name has been correctly changed to **localgov_topics** in my testing.  I have deployed this change to one of our dev sites with an outdated database where field renaming has not happened yet.  That site looks okay just after deployment.  I will confirm tomorrow if it is still unaffected.

Adnan-cds:
> I will confirm tomorrow if it is still unaffected.

Dev site looks good.  Fix seem to be working :)

andybroomfield:
Next step is to get an automated test written for this, if anyone has tips on creating a test with paragraph fields.

andybroomfield:
Putting this back in draft as the tests are passing even if the paragraph fix is removed.
... So its finding out what is causing the underlying deletion and running that directly in the test, as just running cron wont work.

andybroomfield:
I get a test to work (and fail by doing a different type) by running the private function from entity referenece revisions: 
````php
_entity_reference_revisions_orphan_purger_batch_dispatcher('entity_reference_revisions.orphan_purger:deleteOrphansBatchOperation', 'paragraph', []);
````
As the current problem is just paragraphs this fix will work for that.
However all entity referenece revisions fields will also need this fix.
We could just set an exception now if another field type attempts to use this, just saying only paragraphs are supported. 
Really this class should only be called from our update hooks anyway.

Adnan-cds:
> We could just set an exception now if another field type attempts to use this, just saying only paragraphs are supported.

Fine by me :)

andybroomfield:
This is ready for review now @Adnan-cds @ekes

Adnan-cds (approved):
I have finished testing these latest changes on an older copy of the Croydon site where fields were still using the "field_" prefix.  Site looks good.  Tests are also passing.  Code is impressive.  Thank you.

andybroomfield:
I'm going to merge this, though #78 might need to be looked at if its a serious issue.

PR #80, opened 2021-05-26
Fix GitHub CI after 2.0.0 release

stephen-cox:


stephen-cox:
Merged as changes are identical to approved PR here: https://github.com/localgovdrupal/localgov_subsites/pull/42

PR #81, opened 2021-05-26
Forward port of paragraphs field rename bugfix to 2.x

andybroomfield:
Port the fix in #76 to 2.x branch

- Update paragraph tables with references to new field names
- Coding standards fixes
- Add test for paragraph fix on field rename
- Coding standards fix and rename the test
- Fix test to call the entity reference revisions purger
- Add extra if to fieldrenamehelper to detect entity_referenece_revisions
- Throw exception if renaming a non paragraph entity_reference_revisions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,9 @@
         "drupal/redirect": "^1.6",
         "drupal/token": "^1.7"
     },
+    "require-dev": {
+        "drupal/paragraphs": "^1.12"
+    },
     "extra": {
         "enable-patching": true,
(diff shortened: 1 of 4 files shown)


andybroomfield:
If tests on this pass, I'll let @stephen-cox review just to make sure its done correctly.

andybroomfield:
Looks like we need #80 added before this.

stephen-cox:
#80 has been merged, so merging 2.x into this should fix the test run

stephen-cox (approved):
Looks good

PR #84, opened 2021-06-02
adds global variable to turn on/off CSS/JS

### LINKED ISSUE #83: Add Global Variable so we can turn base theme CSS and JS on and off in sub-themes.
I am creating a theme setting for
- Add CSS
- Add JS

This will allow sub-themes that wish to only use the new base theme HTML to turn the CSS and JS off rather than have to overrides lots of libraries and/or load lots of CSS they do not need.

To use this now, you simply need to add the following in any template:

```
{% if localgov_base_use_css %}
  {{ attach_library('localgov_base/teaser') }}
{% endif %}
```

And follow the same approach for JS:

```
{% if localgov_base_use_js %}
  {{ attach_library('localgov_base/teaser-js') }}
{% endif %}
```

Then if a sub-theme chooses to not use our CSS, the CSS library will simply not load.

You can see a video of this approach here: https://www.youtube.com/watch?v=-MW6MObP_44

markconroy:
Closes #83


danchamp (approved):

danchamp:
Tested and working as expected, thanks @markconroy

finnlewis:
@markconroy if we are following protocol, you are now free to merge this in. No objections, it has been approved and tests pass!

Press the button! :)

markconroy:
@finnlewis 

I'm actually going to NOT merge this, and change it to work in reverse, so sites are not broke without this config.

https://github.com/localgovdrupal/localgov_base/issues/70

Issue #30, opened 2020-08-05
Installing permissions for authors and content editors


stephen-cox:
The localgov_roles module installs permissions for the 4 service content types. It doesn't do anything for other content types as we don't know which other content types will be enabled at install time. We need some way of installing these so athours and editors can use the workflow defined in localgov_workflow.

Adnan-cds:
I came across this very issue while setting up the LocalGov Croydon site :(

FAO @finnlewis @stephen-cox @willguv @andybroomfield

willguv:
I expect this will come up in next week's retro - there was extra dev work we needed to do to bridge the gap between LGD and having a production ready site

Could optional modules have bundled workflow, or look for workflow in services (if services is installed)?

ekes:
## Background

Related: 
 * https://github.com/localgovdrupal/localgov_directories/issues/95
 * https://github.com/localgovdrupal/localgov_directories/issues/15

Also for information:
https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/node/node.install#L125
```
  // IMPORTANT: Modules SHOULD NOT automatically grant any user role access
  // permissions in hook_install().
```
although `node` moudle then goes to for the `access content` to anon and authenticated. As does media in media.install. And we then do the same for view permissions in https://github.com/localgovdrupal/localgov_geo/blob/3d5a905ff3ac634401daaa9940ddcd5d031df76c/localgov_geo.install#L14 Again for anon and authenticated.

But the problem then arose that when the site in #95 was configured permission to create Venue content was granted, but permission to create the Geo entity wasn't.

Except for services, as mentioned no other content: guides, step-by-step, directory add permissions for this role.
So the desire is really about the role provided by this module:

https://github.com/localgovdrupal/localgov_directories/issues/95#issuecomment-853002690
> content designers need to be able to create, edit and publish all types of directories content.
> Croydon has these permission now and use them all the time. While I was there, Brighton content designers created everything they could too. (Creating facets wasn't available until you built them, but was always part of the plan).

## Proposal

So it feels like this module (localgov_role) should be assigning the permissions to its role. It of course doesn't know about which permissions are being added. But by maintaining most of the logic in this module it saves repeating it in every single localgov module that wants, and it might make it clearer.

[...]

PR #87, opened 2021-06-07
Add HOOK_localgov_roles_default to install default permissions for modules

ekes:
https://github.com/localgovdrupal/localgov_core/issues/30#issuecomment-855784146

### DIFF
--- a/modules/localgov_roles/src/RolesHelper.php
+++ b/modules/localgov_roles/src/RolesHelper.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Drupal\localgov_roles;
+
+use function call_user_func;
+
+/**
+ * Helper class to for Roles.
+ */
+class RolesHelper {
+
+  /**
+   * Content designer role machine name.
+   */
+  const EDITOR_ROLE = 'localgov_editor';
+
+  /**
+   * Author role machine name.
+   */
+  const AUTHOR_ROLE = 'localgov_author';
+
+  /**
+   * Assign permissions to roles if module has defaults.
+   */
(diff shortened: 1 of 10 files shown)


ekes:
Modules implement hook_localgov_roles_default to define what permissions go to which roles (example https://github.com/localgovdrupal/localgov_core/pull/87/files#diff-9cea1813b6a77cbfb77645210cb786155d4287dd75cab6e669e3e9b0e2801f26R11 ) We could also move the defaults for services too out of the module's config/install and put them in the appropriate services modules. 

The default will be applied which ever way round modules are enabled. This is a one-shot-deal on install, after that permissions can be changed.

modules/localgov_roles/src/RolesHelper.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Drupal\localgov_roles;
+
+use function call_user_func;
```
Adnan-cds: Do we really need ```use function call_user_func```?

Adnan-cds (approved):
I haven't tested the functionality.  But code looks good :)

modules/localgov_roles/src/RolesHelper.php
ekes: Probably not. intelephense added it, like it does other use statements, but not usually for non-namespaced global functions, and I didn't remove it :shrug: I only used call_user_func rather than a variable function name out of readability.

Out of curiosity I had a quick look round some vendor puts a use statement in, most doesn't, drupal core like the rest of vendor sometimes prepends the function with a \ for namespace, sometimes doesn't.

PR #91, opened 2021-06-14
Documentation for 459b0f0 fixes #30.

### LINKED ISSUE #30: Installing permissions for authors and content editors

ekes:
README with module explanation, and documentation of the new hook to set default permissions.

### DIFF
--- a/modules/localgov_roles/README.md
+++ b/modules/localgov_roles/README.md
@@ -1,0 +1,28 @@
+# LocalGov Drupal default roles
+
+Provides the default roles used by LocalGov Drupal:
+
+ * Author (`localgov_author`)
+ * Content Designer (`localgov_editor`)
+
+## Adding default permissions
+
+Modules that wish to provide default permissions for these roles can implement
+`hook_localgov_roles_default()`. An example can be found in the test module
+`tests/localgov_roles_test_one/localgov_roles_test_one.module`.
+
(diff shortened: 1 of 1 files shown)


finnlewis:
I've just been testing this locally, and found schema changes in office_hours module. 

I've created this issue https://github.com/localgovdrupal/localgov_paragraphs/issues/32

And: https://github.com/localgovdrupal/localgov_paragraphs/pull/33/files

andybroomfield (approved):
Not sure what is meant by the title, however I think this is fine to add the documentation around permissions.
(I assume this is for localgov_roles only)

ekes:
The title refers to https://github.com/localgovdrupal/localgov_core/commit/459b0f0d9eaa21d680907ff8fc5a0f5bde3f9968 which added the HOOK_localgov_roles_default to install default permissions for modules (requested in #30) and is the primary reason I wrote the README so that people would know how to implement the hook.

finnlewis:
@ekes @stephen-cox any idea why the tests are still failing here, but passing locally? 

Does the test runner cache old builds at all?

PR #109, opened 2021-07-09
Install and enable the Preview Link and Responsive Preview modules

### LINKED ISSUE #94: Workflows : preview draft content exactly as it will appear
## User story

**As a** content author

**I want** to preview the draft content exactly how it will appear

**So that** I can see how my content looks and feels and reads before sending for review.

## Notes: 

Drupal preview is historically rubbish. Is this still the case? If so, do we need another solution?

https://www.drupal.org/project/preview_link ?

stephen-cox:
Closes #94


andybroomfield:
Should these be added to core?
Localgov core gets installed with almost every LocalGov module, yet a site might not be wanting to use Preview Link and Responsive Preview and they don't actually do anything in this code, they are just brought in as dependencies.

Is this for the project / distribution. If that is the case, then adding them as dependencies to the profile and enabling them there would be the best option. 

See also #111 as I think its time localgov_workflows was made its own module.

finnlewis:
Thanks for your comments @andybroomfield - let's have the discussion on https://github.com/localgovdrupal/localgov_core/issues/111 about modularity and dependencies.

Maybe we should create a new branch for the separation work if that is the direction we agree.

finnlewis (approved):
Looks good to me. @andybroomfield are you OK with is to push forward with these pull requests before considering separate pull request to split out localgov_workflows module?

andybroomfield (approved):
Ok, per resolotuon in #111 that this will get removed when localgov_workflow is its own module.

finnlewis:
Thanks @andybroomfield ! 

@stephen-cox let's merge !

Issue #111, opened 2021-07-09
Break open LocalGov_core submodules into stand alone modules


andybroomfield:
As a lot of extra dependencies are now being added (seeing scheduled transitions) which might not always be required by sites if they are not using the workflow, I think it is time that at least localgov_workflow and localgov_roles where made into their own module and the dependencies added there.

This is important as Localgov_core is required by the majority of Localgov modules, which means installing a lot of modules that end up not being used. Ideally core should be minimal for any shared configuration and dependencies added only.

Alternatively, the dependencies such as scheduled_transitions are added as composer suggestions instead of composer require. This means the core module can be installed for common required code like the title block without having to pull down a lot of dependencies.

finnlewis:
@andybroomfield mentioned this in standup. Would be good to gather other people's thoughts. 
@ekes @stephen-cox @Adnan-cds

Adnan-cds:
As far as I understand, localgov_core's submodules are always needed.  So packaging them as separate modules probably won't do any good as they will be downloaded anyway.

If we really don't want to have lots of hard dependencies for localgov_core, then its code/config will have to be tweaked to make those dependencies optional.  For example, the localgov_directories module has an [optional dependency](https://github.com/localgovdrupal/localgov_directories/blob/3a1b986bee1c6776deac9ca9170e622ef8a82a73/localgov_directories.module#L50) on the localgov_services_navigation module.  *Kind* of like that.

andybroomfield:
Thanks @Adnan-cds 
Just to clairfy, We've never so far needed to enable any of the sub modules in localgov_core (They replicate existing BHCC functionality). At least within the services stack (localgov_services, localgov_guides, localgov_step_by_step) they do not require the enabling of anything within localgov_workflows or localgov_roles and these two modules are the main things I think are candidates for seperation. localgov_media I think does get used by news and subsites so theres a case for that one to stay.

Yes I like the idea of selctive enabling functionlity, though I think that is more suitable for resolving the [inclusion of scheduled transitions for the alert banner module](https://github.com/localgovdrupal/localgov_alert_banner/issues/115#issuecomment-868451479)

ekes:
On a general level I'm in favour of keeping localgov_core to really just that. Absolutely essential functionality that all localgov_ modules need and can just assume exists. Everything else can be switched on optionally, even if it means sometimes having to add a logic check to see what else is enabled. 

[...]

PR #113, opened 2021-07-14
2.1.0 release

stephen-cox:
Splits LocalGov Workflows submodule into it's own module; https://github.com/localgovdrupal/localgov_workflows

Changelog:

#84 @markconroy @danchamp Adds global variable to turn on/off CSS/JS.
#102 @finnlewis @Adnan-cds Relabel localgov_media for consistency.
#111 @andybroomfield @stephen-cox @finnlewis Split localgov_workflows into separate module


composer.json
```diff
@@ -11,7 +11,11 @@
         "drupal/linkit": "^6.0-beta1",
         "drupal/metatag": "^1.14",
         "drupal/pathauto": "^1.8",
+        "drupal/preview_link": "^1.7",
         "drupal/redirect": "^1.6",
+        "drupal/responsive_preview": "^1.0",
+        "drupal/role_delegation": "^1.1",
+        "drupal/scheduled_transitions": "^2.1",
```
finnlewis: @stephen-cox are these the projects we want to remove from localgov_core in favour of https://github.com/localgovdrupal/localgov_workflows/blob/1.x/composer.json#L10 ?

finnlewis (changes requested):

composer.json
andybroomfield: Should we also use [composer suggest ](https://getcomposer.org/doc/04-schema.md#suggest)here to reccomend users install localgov_workflows or will that be covered in the profile update.

stephen-cox: Good catch - PR for removing these here #114

finnlewis (approved):

PR #121, opened 2022-01-26
Correcting package name

andydempster:
Small change but brings this module inline with the others


finnlewis:
Thanks @BrizzleRocker !

PR #130, opened 2022-04-21
Fix/129 localgov media dependencies

ekes:
Just that. It moves the media types and dependencies, and the also their view modes and dependencies into install; and adds explicit dependencies on the required modules.


ekes:
Ooh fun. So why do the configuration items already exist in some test cases.

ekes:
> Ooh fun. So why do the configuration items already exist in some test cases.

Because they are already provided by some profiles:-

```
> find . -name core.entity_form_display.media.image.default.yml
./web/modules/contrib/localgov_core/modules/localgov_media/config/install/core.entity_form_display.media.image.default.yml
./web/core/profiles/standard/config/optional/core.entity_form_display.media.image.default.yml
./web/core/profiles/demo_umami/config/install/core.entity_form_display.media.image.default.yml
```

Is this a classic example of a reason to namespace _all_ _the_ _things_?

ekes:
> Is this a classic example of a reason to namespace all the things?

https://github.com/localgovdrupal/localgov_core/pull/11#issuecomment-641415782=

Bugger. Of course the _only_ profile _they_ ever install in is their own.

ekes:
> Bugger. Of course the only profile they ever install in is their own.

But fine then. This config can live in the profile of anything that's installing localgov_subsites or any other module requiring on this media config.

Nope. Profile configuration is installed last https://www.drupal.org/node/3118908

Add localgov_subsites to the standard profile (or even put the config into `config/install` in another profile with localgov_subsites listed) and it won't install, because clearly the dependency calculation doesn't take into account the config in the profile yet.

ekes:
> Bugger. Of course the only profile they ever install in is their own.

Nope not true https://git.drupalcode.org/project/lightning_media/-/commit/f6e66690dae69271c0ff1c694c5138d1ca3fc710 
They moved everything into config/optional to make it possible to install all standard.

So... is something special done with dependencies or does everything that then requires it also have everything moved to optional?

ekes:
I'm going to close this PR.

There's a question when we get a chance to refactor the paragraphs that depend on media how we do that.

For now for a profile a not very satisfactory work around is to add the _required_ configuration into a _module_ in the profile. Yep. As it's the profile installing a module, even if it lives in the module, that is getting the config/install configuration included early enough and 'overrides' the optional stuff in localgov_media, but is there for the paragraphs already.

Issue #135, opened 2022-05-17
Tests failing because of assigning role permissions in localgov_core/modules/localgov_media/localgov_media.install:17 


ekes:
Testing module that depends on and installs localgov_media.
Even after adding explicit dependencies to filter module (that has the route).
And after moving the config for filter format wysiwyg to install rather than optional (it's really not optional if we need it during the installation) as that's the argument of the route.
But nope not helping.

```
1) Drupal\Tests\localgov_microsites_group\Functional\LoginTest::testLoginRedirect                         
Symfony\Component\Routing\Exception\RouteNotFoundException: Route "entity.filter_format.edit_form" does not exist.
                                                                                                                                                                                                                    
/app/web/core/lib/Drupal/Core/Routing/RouteProvider.php:206                                               
/app/web/core/lib/Drupal/Core/Routing/UrlGenerator.php:426                                                
/app/web/core/lib/Drupal/Core/Routing/UrlGenerator.php:270                                                
/app/web/core/lib/Drupal/Core/Render/MetadataBubblingUrlGenerator.php:105                                                                                                                                           
/app/web/core/lib/Drupal/Core/Url.php:763                                                                 
/app/web/core/modules/filter/src/FilterPermissions.php:56                                                                                                                                                           
/app/web/core/modules/user/src/PermissionHandler.php:153                                                                                                                                                            
/app/web/core/modules/user/src/PermissionHandler.php:112                                                                                                                                                            
/app/web/core/modules/user/src/Entity/Role.php:202                                                        
/app/web/core/lib/Drupal/Core/Config/Entity/ConfigEntityBase.php:319                                                                                                                                                
/app/web/core/modules/user/src/Entity/Role.php:179                                                                                                                                                                  
/app/web/core/lib/Drupal/Core/Entity/EntityStorageBase.php:566                                                                                                                                                      
/app/web/core/lib/Drupal/Core/Entity/EntityStorageBase.php:521                                            
/app/web/core/lib/Drupal/Core/Config/Entity/ConfigEntityStorage.php:253                                   
/app/web/core/lib/Drupal/Core/Entity/EntityBase.php:339                                                                                                                                                             
/app/web/core/lib/Drupal/Core/Config/Entity/ConfigEntityBase.php:607                                                                                                                                                /app/web/modules/contrib/localgov_core/modules/localgov_media/localgov_media.install:17                                                                                                                             
[...]

PR #136, opened 2022-05-17
Avoid error #135 when assigning permissions.

ekes:
Not sure why moving it later works. I still think it must be a
dependency order thing for the wysiwyg filter configuration.


finnlewis (approved):
Looks good to me @ekes and trivial enough in essence. 

We need to get this merged and I guess do a release to help unblock the testing on localgov_microsites_group

finnlewis:
Wanna merge it @ekes ?

PR #137, opened 2022-05-23
release 2.1.4 

finnlewis:


ekes:
Looks fine just github workflow, and the one issue from me that should only make a difference when someone is installing localgov_media and even then shouldn't be noticable (the opposite in fact stop you noticing a rare error).

ekes (approved):

PR #140, opened 2022-07-29
Pin ctools due to update that breaks pauthauto

stephen-cox:
See https://www.drupal.org/project/ctools/issues/3300682

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,7 @@
     "license": "GPL-2.0-or-later",
     "minimum-stability": "dev",
     "require": {
+        "drupal/ctools": "3.9",
         "drupal/field_group": "^3.1",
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Tested locally on a default install of localgov and this fixes the error.

ekes:
https://www.drupal.org/project/ctools/issues/3300682#comment-14636292
Looks like we need to update all the pathauto patterns that get installed.

> To fix your site so that it can use the latest CTools, search your site's configuration for "node_type" in the pathauto.*.yml files and replace it with 'entity_bundle:node';

Although new version of pathauto drops ctools requirement, and ctools is fixing the issue too; but it needs to be updated before D10 anyway.

PR #145, opened 2022-09-14
Add a freestyle crop, image styles and responsive image style, image…

msayoung:
… view mode

### DIFF
--- a/modules/localgov_media/config/optional/core.entity_form_display.media.image.default.yml
+++ b/modules/localgov_media/config/optional/core.entity_form_display.media.image.default.yml
@@ -20,9 +20,11 @@ content:
     settings: {  }
     third_party_settings: {  }
   field_media_image:
+    type: image_widget_crop
+    weight: 1
+    region: content
     settings:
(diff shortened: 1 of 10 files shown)


finnlewis (approved):
Tested locally and works for me on a fresh install with localgov_demo enabled. 

This will only affect fresh installs too. 

Thanks @msayoung !

markconroy (commented):
Checked on microsites and this does what it says on the tin, however I'm wary of the responsive freestyle image style. Just because someone is on a large screen does not mean they will want each freestyle image to be 1200px wide. For example, embedding a headshot of a councillor. you might just want it to be 300px wide.

Is this a concern? If not, I'm happy to approve.

msayoung:
It's true - Do you have any thoughts on how we deal with this @markconroy ?

finnlewis:
From our discussion on the Merge Monday call, I think we're good to go with this @msayoung 

We noted we would follow up on another issue with @markconroy to explore how to give content designers the info they need.

PR #146, opened 2022-09-16
Add media_library_edit to allow media items to be edited after they'v…

msayoung:
…e been uploaded

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,7 @@
         "drupal/field_group": "^3.1",
         "drupal/image_widget_crop": "^2.3",
         "drupal/linkit": "^6.0-beta1",
+        "drupal/media_library_edit": "^2.3",
         "drupal/metatag": "^1.14",
         "drupal/pathauto": "^1.8",
         "drupal/redirect": "^1.6",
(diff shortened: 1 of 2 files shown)


andybroomfield:
Question, should this be added to the profile instead of core?
As its adding something as a standard experince, but there is nothing in localgov_core that would depend on it?
Although I think the other content types will need to be updated to use media_library_edit as the field widget.

Otherwise its a good call, we use it at BHCC as its odd it doesn't do this be default.

msayoung:
Hi @andybroomfield 
There is an issue to get the functionality in to Drupal core, and a patch that we could use. https://www.drupal.org/project/drupal/issues/2985168 
It's been delayed though, pending input from the UX and a11y teams. 

I put this in localgov_core as the localgov_media submodule lives here. But happy to move it if that's better. 

There's another PR for adding this to localgov_paragraphs - and if these two go in, I'll update localgov_news and localgov_events also. I _think_ this should cover it, but there may be more.

finnlewis:
I think we should add it here. The fact that we have submodules means we need to add dependencies for those submodules to our project composer.json. Either that or split out the submodules. So far we've been opting for the former.

finnlewis (approved):
I'm happy with this! Thanks @msayoung and @andybroomfield ! :)

andybroomfield (approved):
As per merge Monday, as used by other modules.

To consider in future if localgov_media should be own module.

PR #150, opened 2022-09-21
Remove dependency on field_formatter_class module

ekes:
Added by a third party setting that does not seem necessary.


msayoung (approved):

finnlewis (approved):
Looks like this is good to merge @ekes :)

PR #152, opened 2022-10-27
Adds a mechanism for subscribers to PageHeaderDisplayEvent to set the cache tags on the PageHeaderBlock.

rupertj:
This is to fix an issue on guide pages. On guide pages \Drupal\localgov_guides\EventSubscriber\PageHeaderSubscriber displays the title and summary of the guide parent page in the PageHeaderBlock, but has no way to update the cache tags on the block to allow the block to update when the parent page is changed.

This PR adds a mechanism for listeners to PageHeaderDisplayEvent to override the block's cache tags if they include data from other entities.

### DIFF
--- a/src/Event/PageHeaderDisplayEvent.php
+++ b/src/Event/PageHeaderDisplayEvent.php
@@ -40,6 +40,13 @@ class PageHeaderDisplayEvent extends Event {
   protected $visibility = TRUE;
 
   /**
+   * Cache tags override.
+   *
+   * @var array|null
+   */
+  protected $cacheTags = NULL;
+
+  /**
    * {@inheritdoc}
    */
   public function __construct($entity) {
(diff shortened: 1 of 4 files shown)


siliconmeadow:
LGTM :-)

finnlewis:
Thanks for this @rupertj 
Hopefully @stephen-cox @ekes @andybroomfield or @Adnan-cds will be able to review, test and approve soon. Maybe for Merge Monday next week, I'll add you to the invite for that in case you can join us.

rupertj:
Thanks @finnlewis. Yeah, please do invite me along on Monday. I've made a follow up PR in localgov_guides (https://github.com/localgovdrupal/localgov_guides/pull/98) to fix the issue using the mechanism this PR adds, so we can go through them both together.

ekes (approved):
For context adding tags for more than just entity page https://github.com/localgovdrupal/localgov_guides/pull/98/commits/d7202f2539306eb9813696bb8211994049ec8233
Can't help feeling there's some other way of doing this. But setting as is, adding cache tags makes complete sense. Test correctly fails without.

Issue #158, opened 2022-12-16
Drupal 10 support


stephen-cox:
Issue to track any changes needed to support Drupal 10

stephen-cox:
Adding non-existent permissions to a role is not allowed. The incorrect permissions are "use text format wysiwyg". 

Caused by https://github.com/localgovdrupal/localgov_core/blob/f857b3e566e4687087bffd7e277fc4a594cefb04/modules/localgov_media/localgov_media.module#L17

I think this is because the relevant config is in optional and so not guaranteed to be installed.

stephen-cox:
@ekes @finnlewis @Adnan-cds @andybroomfield  Having the localgov_media config in `optional` has caused us all sort of issues when trying to use LGD module in other places (Microsites being a prime example). I seem to remember this was done so we don't cause problems when trying to use LGD modules with existing sites. But this is no longer an issue and pretty much everyone is going to be installing LGD with the profile.

Can you think of any reasons we shouldn't move the config in the ;ocalgov_media module into `install` to make it required? If not, I propose doing this as part of the Drupal 10 upgrade work along with a minor / major release of localgov_core.

andybroomfield:
What would be the issue with the config being in optional?
I thought that the config would still be installed if the config is in optional, providing the dependencies are present.

Looking at the config, it is the image styles and media entities that get installed, which one would expect anyway, so that moves it in favour of putting in install.

The counter is that some councils (ahem BHCC) are looking at removing many of the image formats included and replacing with our own as we have different image styles that we are planning to use, though it does take looking through where certian images and media displays are dependant on them.

So I guess we need to decide if we expect (on at least first install) that modules will have the media and image styles avalible, and then pick up issues to make them configurable for those that need to change them.

andybroomfield:
By @ekes in #129

> The localgov_media is all in optional so I think it gets evaluated later. If there is a module depending on these (localgov_homepage_paragraphs, localgov_subsites) depending on how they are installed it can fail

and

> See #130 for a blow-by-blow discovery of why they are in optional, and probably should stay there, and what we should do or not do in the future.

stephen-cox:
Thanks @andybroomfield I knew there should be some discussion about having media config in optional somewhere.

[...]

PR #161, opened 2023-01-25
Drupal 10 support

stephen-cox:
This adds Drupal 10 support. The two major changes that come with this are:

## Now installs CKEditor 5 

CKEditor 4 is currently deprecated and removed in D10. We probably want to check the config for this. Do we perhaps want to move to CKEditor 5 now?

## ~~Media config is now in `install` rather than `optional`~~

~~This hasn't caused any further issues with installing other modules, but should be tested with current install and Microsites as well.~~


finnlewis:
Very cool @stephen-cox !

andybroomfield:
> CKEditor 4 is currently deprecated and removed in D10. We probably want to check the config for this. Do we perhaps want to move to CKEditor 5 now?

What are the potentional issues here? Assume this would install for anyone on new sites, others would still need to upgrade? Is this something that would justify a 3.x branch of core?

andybroomfield:
Also did #158 and the followup get resolved, ie. moving the media into install. Assume it did based on
> This hasn't caused any further issues with installing other modules, but should be tested with current install and Microsites as well.

but be good to confirm.

finnlewis:
Discussed in Merge Monday, continue to test and review this week, not ready for merge yet.

- [ ] test in microsites profile

Putting into draft

stephen-cox:
I have found a  potential issue with moving config in LocalGov Media to install.

Much of this config also exists in the Standard Drupal profile that comes with core. If you try to enable LocalGov Media when using the Standard profile it will fail with an error about config already existing.

Is there a use case for installing LocaGov modules in a site using the Standard profile? If so we're going to have to think of a way around this. For example, keep config in optional and add a hook that installs the config if using the localgov profile.

andybroomfield:
@stephen-cox yes, I don't think we should assume the profile is how everyone installs modules, the goal was that modules can be installed independantly which really could mean installing them in a otherwise standard profile site (I've also been testing modules this way).

Also I thought the arguments around requiring or installing modules in the profile was that profile would set up how the site is intended to use, but those who didn't want could install modules independantly.

As a workaround, could the already in standard media config be moved to optional? or some module use a config/override directory and place config in there?

ekes:
I for one have chatted to people about installing: Directories, Step-By-Step, and Alert Banner outside of the profile. I'd expect Subsites might be another.

ekes:
Is the sticking point on this one the media config location?

ekes (approved):
New installs will get ckeditor 5 works fine. It is available for switching for existing installs.
Tested against an existing d9 localgov test db and nothing adverse. Permissions and ckeditor5 on a new install look good. Tests pass.

Is there some way of tracking where the permissions are getting installed and doing that step-wise? Even maybe not removing them all from here yet till the others are committed?

finnlewis:
Discussing in Merge Monday with Ekes.

Currently we have lots of merge requests for Drupal 10, some approved, but planning to not merge these yet and discuss to review how to get the D10 branches working without breaking the D9 installs.

Suggesting that we either need to:
 - focus on the D9 branches, keeping them working while making releases that start to support D10
 - co-ordinate merge and release across all projects.

Note: we have a Technical Governance meeting on Wednesday where we will focus on confirming the strategy for this.

Ekes suggested listing all modules: 
 - projects that will support D9 and D10
 - projects that will need different branches. 

See https://github.com/localgovdrupal/localgov/issues/480

PR #169, opened 2023-04-07
Adds a separate check for node_preview to entity definition code

ctorgalson:
- Slightly restructures the loop to avoid nesting conditionals and to enable the structure to accommodate other nonstandard cases if any should come up.
- The flow for each iteration is now:
    - exit iteration if $options['type'] is not set;
    - set $entity if $options['type'] is 'entity:*';
    - alternately, set $entity if $options['type'] is 'node_preview' AND its view mode is set to 'full'; 
    - set $this->entity to $entity if a) it's set, and b) it's an instance of EntityInterface;
- Re: 168_pageheaderblock_php_does_not_render_lede_in_node_preview

(Manually) tested for the following conditions on multiple LGD content types:

- node does not exist, preview lede displays correctly
- node already exists, preview lede displays correctly

### DIFF
--- a/src/Plugin/Block/PageHeaderBlock.php
+++ b/src/Plugin/Block/PageHeaderBlock.php
@@ -117,14 +117,28 @@ class PageHeaderBlock extends BlockBase implements ContainerFactoryPluginInterfa
     $this->titleResolver = $title_resolver;
 
     // Find the entity, if any, associated with the current route.
+    //
+    // We consider two cases: (1) type entity:*, and (2) type node_preview with
+    // view_mode_id set to 'full'.
     if (($route = $this->currentRouteMatch->getRouteObject()) && ($parameters = $route->getOption('parameters'))) {
       foreach ($parameters as $name => $options) {
-        if (isset($options['type']) && strpos($options['type'], 'entity:') === 0) {
+        if (!isset($options['type'])) {
+          continue;
+        }
+
+        if (strpos($options['type'], 'entity:') === 0) {
           $entity = $this->currentRouteMatch->getParameter($name);
-          if ($entity instanceof EntityInterface) {
-            $this->entity = $entity;
-            break;
+        }
+        else if ($options['type'] === 'node_preview') {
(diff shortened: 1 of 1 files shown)


ctorgalson:
Assuming this passes the existing tests, I can see it'd be good to test the node preview parts, but I can't see how to do that. If anyone has pointers/examples/documentation, I'd be happy to add to the PR.

ctorgalson:
One further request for assistance: in the project I'm working on, this works for LGD content types _except_ `localgov_guides_page`, and `localgov_step_by_step_overview`. On those content types, the lede **does** appear in the preview, but it's the last saved value instead of the preview content.

Can anyone confirm that this is an issue with this PR? (It could be an issue with my project).

stephen-cox:
@ctorgalson I've just tested this and it works well - thanks.

An example test is testPageHeaderBlockDisplay: https://github.com/localgovdrupal/localgov_core/blob/2.x/tests/src/Functional/PageHeaderBlockTest.php#L47-L82. Extending this test might be a little tricky as you'll need generate a preview link and then login as a user with permissions to view it.

One option would be to write a new test that logs in as a user who can manage content, fill in the fields of the node edit form, click the preview link and then check the lede is visible. We don't have many tests in LGD that do this. I'll post a link if I can find one

ctorgalson:
Thanks @stephen-cox. I'll give that a shot, probably next week.

finnlewis:
Thanks @ctorgalson ! Let us know how you get on. A test would be great, but if this is urgent, @stephen-cox is happy with it as is.

stephen-cox:
@ctorgalson We'll merge this next week at Merge Monday unless you would prefer we hold back until there's a test for this.

ctorgalson:
@stephen-cox Please go ahead when it's convenient (and if you're satisfied with the patch!)

I didn't make any headway on the test, and I'm unable to do contrib work in the next few weeks.

stephen-cox (approved):
This fixes the preview of lede content for me. Let's merge and release at the next Merge Monday (or the reschedule day as the next 2 Mondays are Bank Holidays in the UK).

PR #173, opened 2023-05-11
Install optional config in localgov_media at module install time

### LINKED ISSUE #172: Fix media config installation
Every now and again there are problems with modules dependent on config provided by localgov_media not being installed. This is generally due to the config in localgov_media being in optional and the config in the dependent module being in install.

It looks like a recent change in Drupal (although I haven't tracked this down) means that optional config is installed after all modules have been installed.

This problem recently broke the Drupal 10 install and was solved by installing optional config in localgov_media as the module is being installed: https://github.com/localgovdrupal/localgov_core/pull/161/commits/ccbead2556fee2e8bcc18ca946cae6e3a5513404

I propose bring this change into the distribution now, rather than waiting for Drupal 10 support to be merged.

stephen-cox:
Fixes #172

### DIFF
--- a/modules/localgov_media/localgov_media.module
+++ b/modules/localgov_media/localgov_media.module
@@ -5,6 +5,7 @@
  * Module file for the localgov_media module.
  */
 
+use Drupal\Core\Config\FileStorage;
 use Drupal\user\Entity\Role;
 
 /**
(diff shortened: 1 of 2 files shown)


ekes (approved):
Been wracking my brain for any issues this might cause.
Can't think of one.
So nice. Will this also mean that we can safely fix all the other optional configs without worrying about media too!

Issue #176, opened 2023-05-26
Generator meta tag self promotion.


rupertj:
I ran into this earlier: https://www.viko.co.uk/blog/local-government-cms/

It's an analysis of which CMS platforms UK local government uses. Drupal's doing pretty well, but it made me think - which of those sites are LocalGov and which are regular Drupal?

One of the methods that analysis uses to figure out what powers a site is to check the Generator meta tag. So how about we alter it to say "LocalGov Drupal" instead of just "Drupal"?

rupertj:
I've had a stab at it. With that commit added the result is:
`<meta name="Generator" content="LocalGov Drupal (https://localgovdrupal.org)"/>`

MariosORION:
I like the idea, makes sense to add the flavour of Drupal that the site is built on.

budda:
Great, simple, idea! +1

PR #177, opened 2023-05-26
Changes generator tag to say LocalGov Drupal.

rupertj:
Implements #176.

### DIFF
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -31,3 +31,16 @@ function localgov_core_template_preprocess_default_variables_alter(&$variables) 
     $variables['localgov_base_remove_js'] = TRUE;
   }
 }
+
+
+/**
+ * Implements hook_page_attachments_alter().
+ */
+function localgov_core_page_attachments_alter(array &$attachments) {
+  foreach ($attachments['#attached']['html_head'] as $key => $attachment) {
+    if ($attachment[1] === 'system_meta_generator') {
+      $attachments['#attached']['html_head'][$key][0]['#attributes']['content'] = 'LocalGov Drupal (https://localgovdrupal.org)';
+      break;
+    }
+  }
+}
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -32,7 +32,6 @@ function localgov_core_template_preprocess_default_variables_alter(&$variables) 
   }
 }
 
-
 /**
  * Implements hook_page_attachments_alter().
  */


finnlewis:
Interesting, I like it! 

Any thoughts @ekes @stephen-cox @andybroomfield @Adnan-cds ?

andybroomfield:
What is the reason to use a code to do this, can it not be done in metatag module?

rupertj:
I didn't think of using metatag to do it, TBH.

But I've just tried it and it'll only add a second one, not replace the existing one:
Screenshot 2023-06-05 at 14 14 33

"Hello" is the one I put in with metatag.

finnlewis:
@stephen-cox suggests that this might better live in the localgov profile rather than in localgov_core, which could be installed in a separate site.

ekes:
Notice thunder does it in the profile and then it is in _attachments rather than attachments_alter is that cleaner? https://git.drupalcode.org/project/thunder/-/blob/6.5.x/thunder.profile#L132

finnlewis:
Sounds like moving it to the profile is probably best.

Also discussing with @markconroy  and @stephen-cox - it would be good to keep the 'Drupal 9...' prefix like Thunder do: 

https://git.drupalcode.org/project/thunder/-/blob/6.5.x/thunder.profile#L140

So maybe: 

'Drupal 9 - LocalGov Drupal (https://localgovdrupal.org)';

finnlewis:
Note from Merge Monday: this should be in the profile, not in localgov_core. 

@rupertj fancy moving this to the profile?

stephen-cox (changes requested):
As mentioned by @finnlewis - it's possible for someone to install some of the LocalGov modules separately and localgov_core would likely be installed as a dependency, so I think this should be move to the profile.

finnlewis:
Checking in at Merge Monday - @rupertj you able to take a look at this? 

If not, @alexburrows might be able to 🛩️

rupertj:
I've been meaning to come back and pick this up and move it to the profile as suggested, but haven't had a chance to yet. If @alexburrows (or anyone else) would like to, they'd be very welcome to.

alexburrows:
I'll take this on, it should really go into the localgov profile itself. So I will create a new PR in the profile repo.

andybroomfield:
Closing as we merged the version of this into the localgov profile.
https://github.com/localgovdrupal/localgov/pull/601

PR #178, opened 2023-06-01
Add patch for redirect validation errors

### LINKED ISSUE #551: Intermittent errors when creating redirects
I'm intermittently getting this kind of error when creating redirects:

image

I can't see any errors in the Network tab in Dev Tools, so it looks like this is a problem with some client-side validating.

Refreshing the page a couple of times often helps, but it's quite a nuisance when you're doing a lot of these.

keelanfh:
Fixes https://github.com/localgovdrupal/localgov/issues/551

Thanks to @andybroomfield for suggesting this patch for an annoying issue when creating redirects, which has caused us some grief while migrating content in the last couple of weeks! I've tested the patch on a clean LGD install and it seems to do the job.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,9 @@
     },
     "extra": {
         "enable-patching": true,
+        "patches": {
+            "drupal/redirect": {
+                "Validation issue on adding url redirect: https://www.drupal.org/project/redirect/issues/3057250": "https://www.drupal.org/files/issues/2022-09-01/3057250-53.patch"}},
         "composer-exit-on-patch-failure": true,
         "patchLevel": {
             "drupal/core": "-p2"


andybroomfield (approved):
The patch is clearing the error for me.
I assume that we don't want to check paths via ajax, and that is what this patch is doing?
And I guess including in localgov_core as this module requires (but does not enable) redirect?

keelanfh:
I must admit I haven't reviewed the code in detail, but yes, it looks like that's the case - so the validation is now done only after submitting the form? I think the current code is problematic because it's only running once.

And yes, `localgov_core` because that's where the dependency is declared. Not sure if it's enabled somewhere else.

finnlewis:
@ekes had a quick read of the code and said 'fine' :)

Issue #181, opened 2023-06-09
Drupal 10 and CKEditor 5 issues/differences


keelanfh:
Thought I'd do a bit of testing with CKEditor in our D10 branch.

A few things I observed:

## Height of body field

The height of the body field (in Service pages, perhaps elsewhere too) is very small compared to in CKEditor 4. Actually smaller than the Summary field.

![image](https://github.com/localgovdrupal/localgov_core/assets/19519457/311e14a2-c611-4497-af75-71c2ca6bacc0)

### D9 for comparison:
![image](https://github.com/localgovdrupal/localgov_core/assets/19519457/169b9b74-eb5c-4074-b41b-9ff41dc34b44)

## Possible to e.g. set link styles on non-links

In D9 you cannot set link styles (external link, PDF link, Start button) on non-links. Similarly for some of the other styles ("Green ticks") which should only be set on certain elements. In D10 you can do this, which means you have something styled as an interactive element which actually isn't interactive.

## Possible to set multiple styles on the same element

In D10 it is possible to set multiple conflicting styles on the same element, e.g. here is an item styled simultaneously as Callout primary, Callout teal and Callout carbon. It should only be possible to set one of these styles

![image](https://github.com/localgovdrupal/localgov_core/assets/19519457/66280e26-bcb9-4a5d-8e18-c262ebd9f4ec)

## Styles do not appear in WYSIWYG preview

In D9 there are some nice preview styles in the WYSIWYG to show what things will look like when you publish them. These are missing in D10:

![image](https://github.com/localgovdrupal/localgov_core/assets/19519457/89b0891c-ae9c-4822-acf4-3178250a44ed)

D9 for comparison:

![image](https://github.com/localgovdrupal/localgov_core/assets/19519457/e0f16187-e900-4a84-87ab-f4706307ce3e)

## New "code" and "code block" styles have appeared

Unlikely to be useful for LGD, should probably be removed

![image](https://github.com/localgovdrupal/localgov_core/assets/19519457/83572bde-4f3e-45d9-a7db-ae9093e3f220)

ekes:
Is that https://www.drupal.org/project/drupal/issues/3241295 fixed in 10.1.x or is the min max rows setting not there?

keelanfh:
> Is that https://www.drupal.org/project/drupal/issues/3241295 fixed in 10.1.x or is the min max rows setting not there?

I'll check. Looks like I was on 10.0.9.

justinepocock:
[...]

PR #185, opened 2023-06-29
CKEditor 5 wysiwyg profile changes

### LINKED ISSUE #184: WYSIWYG style "Alert failure" creates alert-primary, not alert-fail
Split from https://github.com/localgovdrupal/localgov_base/issues/460

The WYSIWYG style "Alert failure" adds the `alert-primary` class. 

https://github.com/localgovdrupal/localgov_core/blob/59805fc7da7ec9808c454c82ce0e851a2ef6678a/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml#L65-L66

This is unintuitive for anyone using `localgov_base` because it has an `alert-fail` style defined in it.

https://github.com/localgovdrupal/localgov_base/blob/1.x/css/components/wysiwyg-styles.css#L27-29

keelanfh:
Since this is config, my understanding is that this should not affect existing installs.

We could be concerned about affecting existing installs but I think this is more logical for new ones.

### DIFF
--- a/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
@@ -32,9 +32,6 @@ settings:
       - sourceEditing
       - '|'
       - insertTable
-      - '|'
-      - code
-      - codeBlock
   plugins:
     ckeditor5_heading:
       enabled_headings:
@@ -113,6 +110,9 @@ settings:
       allow_resize: true
     media_media:
       allow_view_mode_override: true
+    linkit_extension:
+      linkit_enabled: true
+      linkit_profile: default
 image_upload:
   status: false
   scheme: public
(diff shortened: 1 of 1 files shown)


keelanfh:
Closes #184

keelanfh:
I've updated this PR with a couple of other changes. This now includes, all for new installs only:

* Changing class applied to Alert failure to `alert-fail` rather than `alert-primary`
* To fix one of the things mentioned in https://github.com/localgovdrupal/localgov_core/issues/181, I've removed the "code" and "code block" styles. We didn't have these styles in CKEditor 4, and I personally think they clutter up the editor.
* I've re-added the default linkit profile here, which went missing in the move to CKEditor 5.
* Removing bootstrap classes as mentioned in https://github.com/localgovdrupal/localgov/issues/582

kiwimind (approved):
Based on the comment https://github.com/localgovdrupal/localgov_core/pull/185#issuecomment-1621326498, this all looks good.
I realise @finnlewis questioned the linkit addition, however if this is replacing functionality that went missing, that feels like another conversation.

PR #190, opened 2023-10-14
fix: add missing image_loading attribute to config

millnut:
This fixes the following deprecation error;

```
The responsive image loading attribute update for "X" is deprecated in drupal:10.1.0 and is removed from drupal:11.0.0. Configuration should be updated to accommodate the changes described at https://www.drupal.org/node/3279032.
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,10 @@ name: Test localgovdrupal/localgov_core drupal-module
 on:
   push:
     branches:
-      - '3.x'
+      - '2.x'
   pull_request:
     branches:
-      - '3.x'
+      - '2.x'
   workflow_dispatch:
 
 env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         include:
           - localgov-version: '2.x'
-            drupal-version: '~9.4'
+            drupal-version: '~9.5'
             php-version: '8.1'
           - localgov-version: '3.x'
             drupal-version: '~10.0'
@@ -107,7 +107,7 @@ jobs:
       matrix:
         include:
           - localgov-version: '2.x'
-            drupal-version: '~9.4'
+            drupal-version: '~9.5'
             php-version: '8.1'
           - localgov-version: '3.x'
             drupal-version: '~10.0'
@@ -143,7 +143,7 @@ jobs:
       matrix:
         include:
           - localgov-version: '2.x'
-            drupal-version: '~9.4'
+            drupal-version: '~9.5'
             php-version: '8.1'
           - localgov-version: '3.x'
             drupal-version: '~10.0'
@@ -178,7 +178,7 @@ jobs:
       matrix:
         include:
           - localgov-version: '2.x'
-            drupal-version: '~9.4'
+            drupal-version: '~9.5'
             php-version: '8.1'
           - localgov-version: '3.x'
             drupal-version: '~10.0'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,12 +21,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - localgov-version: '3.x'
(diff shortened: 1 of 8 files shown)


millnut:
Not too sure why Drupal 9 is having issues, I've increased from 9.4 to 9.5 to rule that out but it still errors

millnut:
Ok doing some digging [Drupal 9.5.x](https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/responsive_image/config/schema/responsive_image.schema.yml) does not support `image_loading` on the `responsive_image` schema, hence the error.

[Drupal 10.1.x](https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/modules/responsive_image/config/schema/responsive_image.schema.yml?ref_type=heads#L70) however does support it, added in [Drupal 10.1.0](https://www.drupal.org/node/3279032).

How is best to handle this, given Drupal 9 is end of life in November 2023 and Drupal 10.0.x will receive security coverage until December 2023; or do we create a 3.x for Drupal 10.1.x onwards and leave 2.x as Drupal 9.5-10.0.x

andybroomfield:
Is there a way to add the extra attribute in a hook_install for now?
Otherwise we might have to make a 3.x branch for core to either make it Drupal 10 only setting at 10.1 or forward compatible to Drupal 11. It depends on others feel about making 2.x drop Drupal 9 support.
It is just a warning at the moment, so nothing essential, though it would be good to have lazy loading on responsive images (BHCC implemented a theme hack to acomplish this).

finnlewis:
Discussing in Merge Monday.

Let's go for a 3.x branch!

Then we'll want to support minimum core version 10.1

finnlewis:
Leaving this with @millnut to update the pull request.

We'll then aim for a 2.x-beta release

millnut:
Hi @finnlewis PR updated, I guess the failure is due to `localgov_core` 3.x-dev not being required in `localgovdrupal/localgov`. If it's another reason let me know and I'll fix it.

andybroomfield:
The issue with creating a 3.x branch and release is that most modules will require the 2.x version, so they will either have to be updated to allow both versions of localgov_core, or we will need to do 3.x releases to support the new version of localgov_core.

stephen-cox:
LGD Core is used by just about all LGD modules. Having a 3.x branch of this will force us to create 3.x D10 only releases for all  dependent modules that don't already have a 3.x branch; so things like Guides, Step By Steps, Subsites, etc.

If there's a way to fix this with a `hook_install` or similar this is going to be a lot less work.

Where's the deprecation error being generated? Can we mark it as ignored for now and fix it later?

andybroomfield:
@stephen-cox It happened when I was running BHCC tests and this came up in the logs for anything that had a dependency on localgov_services https://github.com/localgovdrupal/localgov_core/issues/191

finnlewis:
Discussing again in Merge Monday.

@andybroomfield suggests waiting for a couple of days until D9 is out of support.

We can then do a minor release which drops support for D9.

No objections to that approach from those on the call.

andybroomfield:
We would explicity then add the minimum supported drupal/core version to composer.json.
The rational here is that if we were to do a 3.x release of localgov_core, we would then need to make a 3.x release of every module that now requires localgov_core, which is almost all modules.
As precedent we have dropped support for unsupported versions of Drupal without a major release before, eg. https://github.com/localgovdrupal/localgov_alert_banner/releases/tag/1.6.0
There is no actual breaking change, but Drupal core less than 10.1 won't be able to install the new config for responsive images.
We should wait until profile as a full 3.x release first though and for 10.0 to go out of support, so this is actully December.

stephen-cox:
from Merge Tuesday. Drupal 10.0 is no longer supported so we can pick this up again.

Will drop support for version 10.0 so might need a minor release to make it clear we've dropped support.

andybroomfield:
Also `drupal/core: ^10.1` should be added to composer.json to prevent accidental upgrade on an unsupported Drupal version.

finnlewis (approved):
Woop woop! Thanks @millnut

PR #193, opened 2023-11-02
Add stacked header subtitle as new page header variable

### LINKED ISSUE #192: Support adding a subtitle inside the h1 in page header block
Ref to 
- https://github.com/localgovdrupal/localgov_guides/issues/122
- https://github.com/localgovdrupal/localgov_guides/pull/133
- https://github.com/localgovdrupal/localgov_base/pull/492
- https://github.com/localgovdrupal/localgov_publications/blob/1.x/src/Plugin/Block/PublicationPageHeaderBlock.php

There seems to be several ways of trying to get a pattern of having a subtitle that is not the summary to be inside the H1 or look like it is.
It would be better to have an official way inside the page header block, with a setSubtitle() method. This would allow consistent theming to be applied for all subtitles, whilst retaining the ability for modules that wanted to send a render array to replace the title being able to do so.

andybroomfield:
Fix #192

This allows page subscribers to set a subtitle which will be stacked inside the h1.

### DIFF
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -13,6 +13,7 @@ function localgov_core_theme($existing, $type, $theme, $path) {
     'localgov_page_header_block' => [
       'variables' => [
         'title' => '',
+        'subtitle' => NULL,
         'lede' => NULL,
       ],
       'render element' => 'block',
--- a/src/Event/PageHeaderDisplayEvent.php
+++ b/src/Event/PageHeaderDisplayEvent.php
@@ -33,6 +33,13 @@ class PageHeaderDisplayEvent extends Event {
   protected $title = NULL;
 
   /**
+   * The page sub title override.
+   *
+   * @var array|string|null
+   */
+  protected $subTitle = NULL;
+
+  /**
    * Should the page header block be displayed?
    *
    * @var bool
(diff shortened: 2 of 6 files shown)


ctorgalson:
Coming over from the [PR 133 on `localgov_guides`](https://github.com/localgovdrupal/localgov_guides/pull/133). 

I like this (though I would quibble pointlessly about that subtitle markup :grinning:). I have two additional suggestions:

1. add another variable, `node`, to the module's `hook_theme` implementation for this block, then populate it with the node object. This will make it simple, in themes, to override the rendering of the block according to content type (and according to the node content itself),
2. add template suggestions for the _content type_ to the block template (maybe using `EntityTypeBundleInfo()` in `hook_theme_suggestions_HOOK()`...not sure if we would want to limit this to LGD content types, or just catch them all),

The fact that what might be relatively simple things to handle in themes spawn issues like [consider stacked h1 pattern for guides](https://github.com/localgovdrupal/localgov_guides/issues/122), and that (as you pointed out) `localgov_publications` uses a _different_ block suggest to me that the core block could be more helpful specifically w/r/t the current content type.

andybroomfield:
Thanks for the suggestions @ctorgalson. I'll look into adding the entity, thats actully one thing to check over as it should be the case the page title block can be used with any entitiy and not just node, and in fact even stand alone controllers (we use it at BHCC to modify search to include the search terms).
However getting the current node shouldn't require getting the route path, so if its avaliiable we should provide it, and perhaps provide a way for other modules to set which entitiy is avalible.

andybroomfield:
I'm marking this as ready for review, although this is just the adding of the subtitle.
@ctorgalson I think we should add the entity to the template and template suggestions as a new PR as that is adding some new functionality and needs some thought on implementation and tests. This should be sufficient to then cover most use cases for the page header (title with lede, title with subtitle inside h1).

markconroy:
This looks good to me.

markconroy (approved):

ctorgalson:
It looks good to me too :+1:

Issue #211, opened 2024-04-17
Installing blocks


rupertj:
The way we currently install blocks doesn't seem to be totally reliable. I'm proposing we add a mechanism to localgov_core to add them reliably.

The problem:
LGD modules that install blocks currently use config to do it - typically with two files per block: one for localgov_base and one for localgov_scarfolk. This works fine for those themes in our tests, and for new installs of a site that are yet to switch to their own theme. Themes that inherit from localgov_base also inherit its block placements when they don't have their own blocks, but themes that inherit from localgov_base and have their own blocks placed don't inherit blocks. (See https://www.drupal.org/docs/develop/theming-drupal/creating-sub-themes#s-inheriting-block-placement).

Hence if you add a module to an existing site that has a custom theme that inherits from localgov_base, the newly installed module's blocks will not be placed.

An initial solution:
As we've run into this issue a few times on localgov_publications, I've fixed it there by placing the blocks in hook_install(), into the site's active theme and the two LGD themes. This works, but if we want to fix this same issue in other modules, we'll end up duplicating that code.

What I'm proposing here:
We put a modified version of the code that's currently placing blocks in publications in a new sub module of localgov_core, which we'll call localgov_default_blocks.

If a module wishes to place default blocks, it declares a dependency on localgov_default_blocks and implements a hook to define the blocks. The contents of this hook will look like what's currently in the function localgov_publications_install_block_definitions() here: https://github.com/localgovdrupal/localgov_publications/blob/157554c463c8d70dbb190340cf4b79a6352c0180/localgov_publications.install#L99 . localgov_default_blocks will work much like the existing localgov_roles module does. It'll implement hook_modules_installed, call the hook to get the default block definitions and then set those blocks up. The code for this'll look a lot like the code in localgov_publications_install_blocks(): https://github.com/localgovdrupal/localgov_publications/blob/157554c463c8d70dbb190340cf4b79a6352c0180/localgov_publications.install#L72.

Questions / comments / feedback please :)

rupertj:
Andy B commented in the slack thread:

> I'd say its fine to try this for blocks in a hook install, just be careful about placing them in the correct region, maybe sitck to onces that follow localgov_base conventions (sidebar_first, sidebar_second) etc, though that still means those on alternative themes have to place them themselves it's no different to current situation.

Which is an excellent point. The code in publications doesn't currently check that the region we're putting blocks into exists in themes, but a central solution should.

rupertj:
[...]

PR #212, opened 2024-04-26
Installing default blocks

rupertj:
Adds a feature that installs default blocks for other LGD modules.

Implements the ideas discussed in #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,22 @@ off by adding this to your site's settings.php file:
 ```php
 $config['localgov_core.settings']['install_default_blocks'] = FALSE;
 ```
+
+If you're a module maintainer and would like to use this feature, create a file 
+in your module at config/localgov/block.description.yml. The description part of
+the filename can be anything you like.
+
+In that file, place the exported config yaml for a single block, and remove the 
+following keys:
+* uuid
+* id
+* theme
+
+The default block installer will read the file, and create an instance of the 
+block in the current active theme, along with localgov_base and 
+localgov_scarfolk, if they exist and are enabled. An id for each instance will
+be generated from the combination of theme and block plugin name.
+
+Using this feature lets your blocks appear automatically in the right place in
+existing localgov sites with custom themes. It also saves you having to manage
+multiple block config files for localgov_base and localgov_scarfolk.
--- a/README.md
+++ b/README.md
@@ -1,3 +1,14 @@
 # LocalGov Drupal core functionality
 
 LocalGov Drupal Core module, for helper functions and core dependencies.
+
+## Default blocks
+This module contains a mechanism that will place default blocks into your site's
+active theme when other localgov modules are installed. This is indended to 
+reduce the work that site owners need to do when installing new features 
+provided by localgov modules. If you don't want this to happen, you can turn it 
+off by adding this to your site's settings.php file:
+
+```php
+$config['localgov_core.settings']['install_default_blocks'] = FALSE;
+```
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ LocalGov Drupal Core module, for helper functions and core dependencies.
 
 ## Default blocks
 This module contains a mechanism that will place default blocks into your site's
-active theme when other localgov modules are installed. This is indended to 
+active theme when other localgov modules are installed. This is intended to
 reduce the work that site owners need to do when installing new features 
 provided by localgov modules. If you don't want this to happen, you can turn it 
 off by adding this to your site's settings.php file:
--- a/src/Service/DefaultBlockInstaller.php
+++ b/src/Service/DefaultBlockInstaller.php
@@ -146,9 +146,22 @@ class DefaultBlockInstaller {
 
   /**
    * Replace characters that aren't allowed in config IDs.
+   *
+   * This is partly based on
+   * \Drupal\Core\Block\BlockPluginTrait::getMachineNameSuggestion().
    */
   protected function sanitiseId(string $id): string {
-    return preg_replace('/[^a-z0-9\._]/', '_', strtolower($id));
+
+    // Shift to lower case.
+    $id = mb_strtolower($id);
+
+    // Limit to alphanumeric chars, dot and underscore.
+    $id = preg_replace('@[^a-z0-9_.]+@', '_', $id);
+
+    // Remove non-alphanumeric chars from the beginning and end of the id.
+    $id = preg_replace('@^([^a-z0-9]+)|([^a-z0-9]+)$@', '', $id);
+
+    return $id;
   }
 
   /**
--- a/src/Service/DefaultBlockInstaller.php
+++ b/src/Service/DefaultBlockInstaller.php
@@ -6,6 +6,7 @@ use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Extension\ThemeHandlerInterface;
 use Drupal\Core\File\FileSystemInterface;
+use Drupal\Core\Installer\InstallerKernel;
 use Drupal\Core\Theme\ThemeManagerInterface;
 use Symfony\Component\Yaml\Yaml;
 
@@ -99,17 +100,22 @@ class DefaultBlockInstaller {
 
     $themes = ['localgov_base', 'localgov_scarfolk'];
 
-    // Don't try to use themes that don't exist.
-    foreach ($themes as $i => $theme) {
-      if (!$this->themeHandler->themeExists($theme)) {
-        unset($themes[$i]);
+    // Check if the site is being installed. If it is, there's no active theme,
+    // and we can't check if themes exist. We'll assume that localgov_base and
+    // localgov_scarfolk are present for new installs.
+    if (!InstallerKernel::installationAttempted()) {
+      $activeTheme = $this->themeManager->getActiveTheme()->getName();
+
+      if (!in_array($activeTheme, $themes)) {
+        $themes[] = $activeTheme;
       }
-    }
 
-    $activeTheme = $this->themeManager->getActiveTheme()->getName();
-
-    if (!in_array($activeTheme, $themes)) {
-      $themes[] = $activeTheme;
+      // Don't try to use themes that don't exist.
+      foreach ($themes as $i => $theme) {
+        if (!$this->themeHandler->themeExists($theme)) {
+          unset($themes[$i]);
+        }
+      }
     }
 
     return $themes;
@@ -127,8 +133,12 @@ class DefaultBlockInstaller {
     foreach ($this->targetThemes() as $theme) {
       foreach ($blocks as $block) {
 
-        if (!$this->themeHasRegion($theme, $block['region'])) {
-          continue;
+        // If we're not in the installer, verify that the theme we're using has
+        // the requested region.
+        if (!InstallerKernel::installationAttempted()) {
+          if (!$this->themeHasRegion($theme, $block['region'])) {
+            continue;
+          }
         }
 
         $block['id'] = $this->sanitiseId($theme . '_' . $block['plugin']);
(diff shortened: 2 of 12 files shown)


rupertj:
There's a few bits of configuration that I want to add to this, but it works. Please take a look and tell me what you think!

I'd recommend starting with the tests: They include a module called localgov_core_default_blocks_test which does nothing but provide a custom block plugin and some configuration to place it using the new mechanism.

andybroomfield:
Tested by installing a BHCC custom module with a block using the new config/localgov block tecnique

Installs on localgov_scarfolk

Screenshot 2024-04-30 at 11 00 36 AM

and on gov.uk theme

Screenshot 2024-04-30 at 11 00 02 AM

I also tested this using the footer region and the content_bottom region, the later missing from govuk theme and verified it is not placed anywhere and didn't see any error messages. (Maybe it is worth adding a message saying the block still needs to be placed). Perhaps a future enchancement could be that the region key is replaced with a regions array of acceptable regions in order?

I'm happy with this approach, though I have a few questions.

I originally tried to test with Olivero, but that theme uses slightly different names for it's regions (content_below instead of content_bottom). This rasises the issue of wanting to provide default blocks for themes but in different regions, which we could do using the existing method of placing it in config/optional.

I have a question on what to do about Alert banner, which currently does not depend on core currently. Switching to this method would mean that if users did want to use localgov_base and localgov_alert_banner they would also need core to get the default block placement or could I continue to provide one in config/optional? Would that mean two blocks end up getting placed if localgov_core is present? I'm thinking the case that is I provided default blocks under /optional for Olivero and that happens to be the active theme. Maybe a way of either detecting or a module declaring that it will handle installing for that theme itself so this service can skip it.

I also notice that it only works with the active theme, so doesn't effect the admin theme (which is good!). Some site (BHCC) have multiple front end themes where default block placement could be useful, or could also be a hinderance depending on perspective. 

Noting there are `@todo` statements on both those issues above, so it's something we can come back to if we need to move this forward. I'm tempted to actully say it should only target the active theme and so module continue as now with providing default blocks and we skip over the localgov_base / scarfolk.

rupertj:
@andybroomfield I'd like it if people didn't have to do both the existing method of using optional config *and* the new one. We could support your use case of install into olivero (or another theme with different regions) by using your idea about an array of themes rather than a single value. So if you wanted a block in content_below on olivero, or content_bottom on a localgov theme, you'l list both, and it'd check the regions in order and install into all the localgov themes, plus olivero if it's set as the active theme.

For alert banner, I'd suggest that you add the yml files for this new mechanism and remove the existing config. There's currently only config there for base and scarfolk, and I have a strong suspicion that anyone who's using one of those themes will have localgov_core installed anyway. Anyone who's got a subtheme of base probably won't get the blocks anyway using the current methodology, so nothing's been lost, and at least if they have a subtheme of base and localgov_core installed, they will get them.

finnlewis:
Discussing this today in Merge Tuesday, general enthusiasm for the approach. 

@Adnan-cds also mentioned the possibility of listing multiple potential regions a block might want to be in.

It'd be good to test this more widely and get this in soon.

finnlewis:
@rupertj I'm testing this, finally!

finnlewis:
I've tested the functionality, and confirmed that it works.

block config placed in /config/localgov does get installed (as long as it's extension is .yml not .yaml :smile: )

Could do with another review of the code if possible, but I've stepped through it with xdebug and it all makes sense to me.

@andybroomfield you happy with the code?

src/Service/DefaultBlockInstaller.php
```diff
@@ -144,6 +144,13 @@ public function install(string $module): void {
     }
   }
 
+  /**
+   * Replace characters that aren't allowed in config IDs.
+   */
+  protected function sanitiseID(string $id): string {
+    return preg_replace('/[^a-z0-9\._]/', '_', strtolower($id));
```
ekes: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Block%21BlockPluginTrait.php/function/BlockPluginTrait%3A%3AgetMachineNameSuggestion/11.x

Also removes not alphanumerical chars from the end, and uses mb_strtolower. I'm assuming we can't just use the trait here, but, probably should be doing the same thing?

rupertj: Good find! I've made it do the same thing, and added a comment referencing that method.

README.md
```diff
@@ -1,3 +1,14 @@
 # LocalGov Drupal core functionality
 
 LocalGov Drupal Core module, for helper functions and core dependencies.
+
+## Default blocks
+This module contains a mechanism that will place default blocks into your site's
+active theme when other localgov modules are installed. This is indended to
```
stephen-cox: Typo /indended/intended/.

stephen-cox (commented):
I may have found an issue with this. To test I moved the base block config for LGD services from the `optional` dir into a `localgov` dir, removed the theme settings from the YAML and deleted the block config for the scarfolk theme.

Site installs fail with the error `The theme core does not exist.`

It looks like when the install hook is running neither the LGD base or Scarfolk themes are enabled and the DeffaultBlockInstaller::targetThemes() method is returning 'core' as a theme.

Am I missing something in the testing here?

src/Service/DefaultBlockInstaller.php
ekes: Nice one.

README.md
rupertj: No, you're not missing anything. I hadn't thought to test this in the installer, to be honest. Good find! I can recreate this locally, so I'll sort it out now.

rupertj: So it turns out 'core' is a kind of fake theme that Drupal will return if there's no other themes: https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Theme/ThemeInitialization.php?ref_type=heads#L99

rupertj: So to fix this, I've added a couple of special cases that check if we're in the installer. One to stop trying to load the active theme, as that's not a thing in the installer, and skip the check that the themes exist, because none of them do yet.

The other skips the check that the theme has the requested regions, as we can't load the theme info to check that either.

I think doing it this way is fine for the installer case, as in practice it'll be mostly be people who are using the default project, where we know what exists and what doesn't.

ekes: If you know that get active theme is going to return `core` in all cases where there isn't an active real theme, is it not better to check the active theme, and if it is `core` bail out? Just in case it happens on some other occasion too?

rupertj: Which reminds me.. The other thing I did was move the check for themes existing below the code that finds the current default. (It was the other way round before). That should handle that case. Although I suspect the only time you'll ever have no theme while installing a module is going to be the installer.

ekes: I'm going to guess it's only the installer... but... maybe some new sync' recipe or something that doesn't use the installer but something else... I don't know. It feels like it is an unnecessary assumption if it can bail out on the fact that the theme isn't there no matter what?

rupertj: Well, if such an installer-like thing does exist in the future, the code as-is will add 'core' (or whatever) to the list of themes, then the next block of code checks all the themes exist, at which point, it'll be taken out of the list. So I think we're good for the moment.

rupertj: Ok, I've changed my mind. Latest commit wraps the call to $this->themeManager->getActiveTheme()->getName(); in a new method called getActiveThemeName() and returns null if the theme name is 'core' to indicate that there's no real active theme.

ekes: :-)

ekes:
I think you're concern @stephen-cox is also fixed now?

stephen-cox:
Unfortunately, I've found another couple of problems with this.

The block installer doesn't check dependencies so it can try to install blocks before the code defining them has been enabled, which displays a warning. This is not a serious issue though, it just means we have to be sure all dependencies are in place in a module that contains the block config.

More of a problem is how Drupal handles enabling blocks in a theme before the theme is installed. It looks like blocks are added to the first theme region and disabled. So the blocks don't show without some manually intervention.

image

For a site install it would be better to have the block installer triggered after the themes are installed, but that's going to mean overhauling how this is triggered.

It would be nice if there was a hook for when site installs are finished, but this doesn't look to exist (https://www.drupal.org/project/drupal/issues/2924549).

rupertj:
Ah, that's annoying. Thanks for the further testing though.

I could add an install step to the end of the localgov profile to set up the blocks?

stephen-cox:
> I could add an install step to the end of the localgov profile to set up the blocks?

Assuming this is going to be used as with site installs (which I assume it will be if we include Publications) then I think this sounds sensible.

Maybe only work for themes that are installed (ignoring core) and then provide a way for projects and themes to rerun the block installer - or something like that.

markconroy:
@rupertj given the general enthusiasm for this approach, and given that we have another issue that is blocked pending this getting merged, would you be able to get the last couple of items completed on this?

Would be great to have it merged. 

Thanks.

rupertj:
I've got this working and made a new issue for it against the profile: https://github.com/localgovdrupal/localgov/issues/750

I'll need to make an MR there to add the new install step and update the MR here to use it.

rupertj:
So I've implemented the install step idea to make this work. To test this, you'll need:

* localgov_core with this MR applied.
* The localgov profile with MR #751 applied.
* A module altered to use the new default blocks config mechanism. There's details on how to make use of it in the MR for this module now, but if you'd like a specific example, here's one for the Services CTA block: Go to the directory localgov_services is installed in and run these commands:
  * rm config/optional/block.block.localgov_servicescalltoaction_scarfolk.yml
  * mkdir config/localgov
  * mv config/optional/block.block.localgov_servicescalltoaction_base.yml config/localgov/block.localgov_services_call_to_action.yml
  * Edit config/localgov/block.localgov_services_call_to_action.yml and remove the id and theme keys.

Now reinstall the site and you should find the CTA block in place as expected.

rupertj:
The instructions above still apply, but for a more streamlined way to test this, I've added a branch to localgov_publications with the modifications to use this new mechanism. To test this:

### Setup
Add this to your composer.json to install the required MRs:
```
"localgovdrupal/localgov": "dev-feature/3.x/750-add-install-step as 3.0.9",
"localgovdrupal/localgov_core": "dev-feature/2.x/211-installing-blocks as 2.13.6",
"localgovdrupal/localgov_publications": "dev-feature/1.x/new-default-block-mechanism as 1.0.5",
```
Run composer update.

### Case 1: module installed after the site is installed.
1. Install localgov.
2. Enable localgov_publications
3. Log in and check the Sidebar first region contains “Publication navigation” and “On this page” blocks.

### Case 2: module installed during site install.
1. Edit web/profiles/contrib/localgov/localgov.info.yml and add: `localgov_publications:localgov_publications`
3. Install localgov.
4. Log in and check the Sidebar first region contains “Publication navigation” and “On this page” blocks.

stephen-cox (approved):
Looking good @rupertj - this works for me.

I would like @ekes opinion on the changes here and in localgovdrupal/localgov#751 but otherwise happy to approve.

src/Service/DefaultBlockInstaller.php
```diff
@@ -0,0 +1,185 @@
...
+   */
+  protected function themeHasRegion(string $theme, string $region): bool {
+    return in_array($region, $this->themeRegions($theme));
+  }
+
+  /**
+   * Gets the regions for the given theme.
+   */
+  protected function themeRegions(string $theme): array {
+    if (!isset($this->themeRegions[$theme])) {
```
Polynya, line 172: getTheme() throws an exception if the theme does not exist so check first, e.g.: 
```
if (!$this->themeHandler->themeExists($theme)) {
  $this->themeRegions[$theme] = [];
  return [];
}

rupertj: The list of themes that's used comes from DefaultBlockInstaller::targetThemes(), which already checks the themes exist.

Polynya: oh yeah

PR #214, opened 2024-05-04
fix: strict types support

### LINKED ISSUE #210: Strict types support
With Drupal core moving towards strict types, [more info](https://github.com/localgovdrupal/localgov/issues/685) I've added a test pipeline in preparation https://github.com/localgovdrupal/localgov_project/pull/153 which has highlighted some failures which we should fix.

```
------ ---------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_core/modules/localgov_media/localgov_media.module  
 ------ ---------------------------------------------------------------------------- 
  16     Call to function in_array() requires parameter #3 to be set.                
 ------ ---------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_core/modules/localgov_roles/tests/src/Kernel/ModuleRolesTest.php  
 ------ ------------------------------------------------------------------------------------------- 
  32     Call to method Drupal\KernelTests\KernelTestBase::setUp() with                             
         incorrect case: setup                                                                      
 ------ ------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_core/tests/src/Kernel/FieldRenameHelperTest.php  
 ------ -------------------------------------------------------------------------- 
  53     Call to method Drupal\KernelTests\KernelTestBase::setUp() with            
         incorrect case: setup                                                     
  178    Call to function in_array() requires parameter #3 to be set.              
  179    Call to function in_array() requires parameter #3 to be set.              
  184    Call to function in_array() requires parameter #3 to be set.              
  185    Call to function in_array() requires parameter #3 to be set.              
 ------ -------------------------------------------------------------------------- 
```

millnut:
Fix #210

Fixes strict type issues highlighted in the above issue.

### DIFF
--- a/modules/localgov_media/localgov_media.module
+++ b/modules/localgov_media/localgov_media.module
@@ -13,7 +13,7 @@ use Drupal\user\Entity\Role;
  * Implements hook_modules_installed().
  */
 function localgov_media_modules_installed($modules, $is_syncing) {
-  if (!$is_syncing && in_array('localgov_media', $modules)) {
+  if (!$is_syncing && in_array('localgov_media', $modules, TRUE)) {
 
     // Install all optional config now rather than deferring it to later.
     $config_path = \Drupal::service('extension.list.module')->getPath('localgov_media') . '/config/optional';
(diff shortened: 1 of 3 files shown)


ekes (approved):
Got to say I think strict type checking on string module names is all a bit much, but if that's what we gotta do.

Also forgot completely that method names weren't case sensitive!

Issue #218, opened 2024-05-22
Set error status messages in localgov_core instead of the localgov_base theme


chriswales95:
This partly relates to the another issue in localgov_base https://github.com/localgovdrupal/localgov_base/issues/535

Currently, the status-messages for localgov themes are set in the base theme templates page--403.html.twig and page--404.html.twig. This makes it harder to customise those pages and isn't the best place for those messages to be set.

Drupal uses the Http4xxController to set the messages that are the default messages that come through that the localgov_base theme hides by using the hard-coded messages in the template. I think we'd be best overriding those messages in the controller instead. 

This means if a user wants to have a custom 404 or 403, they would have full control and wouldn't have to live with the hard-coded messages or supply their own theme.

chriswales95:
I will open a PR for this.

chriswales95:
Closed as discussed in the linked PR.

PR #219, opened 2024-05-22
Override system 400 messages

chriswales95:
Overrides 404 and 403 system messages which have previously been hard-coded in localgov_base.

This relates to issue #218.

### DIFF
--- a/src/Controller/Http4xxController.php
+++ b/src/Controller/Http4xxController.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Drupal\localgov_core\Controller;
+
+use Drupal\Core\Controller\ControllerBase;
+
+/**
+ * Returns responses for LocalGov Core routes.
+ */
+class Http4xxController extends ControllerBase
+{
+
+  /**
+   * The default 403 content.
+   *
+   * @return array
+   *   A render array containing the message to display for 403 pages.
+   */
+  public function on403()
+  {
+    return [
+        'message' => [
+            '#type' => 'html_tag',
+            '#tag' => 'p',
+            '#value' => $this->t('You are not authorised to access this page.')
+        ],
+        'login_link' => [
+            '#type' => 'html_tag',
+            '#tag' => 'a',
+            '#attributes'=> [
+                'href'=> '/user/login'
+            ],
+            '#value' => $this->t('Login?')
+        ]
+    ];
+  }
+
+  /**
+   * The default 404 content.
+   *
+   * @return array
+   *   A render array containing the message to display for 404 pages.
+   */
+  public function on404()
+  {
+    return [
+        'message' => [
+            '#type' => 'html_tag',
+            '#tag' => 'p',
+            '#value' => $this->t('The page you are requesting cannot be found.')
+        ],
+    ];
+  }
+
+}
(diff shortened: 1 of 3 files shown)


src/Controller/Http4xxController.php
```diff
@@ -0,0 +1,55 @@
...
+            '#tag' => 'p',
+            '#value' => $this->t('You are not authorised to access this page.')
+        ],
+        'login_link' => [
+            '#type' => 'html_tag',
+            '#tag' => 'a',
+            '#attributes'=> [
+                'href'=> '/user/login'
+            ],
+            '#value' => $this->t('Login?')
```
andybroomfield, line 33: When we added a login link to our 403 pages, we found customers trying to login with their MyAccount details so I'd suggest not including this.

andybroomfield:
If a site has set custom 404 and 403 pages in /admin/config/system/site-information would this override them?
I wouldn't want to stop our [404 page](https://www.brighton-hove.gov.uk/404) working.

src/Controller/Http4xxController.php
chriswales95: That's true. 

However it's just what's in the hard-coded templates at the moment: https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/layout/status-pages/page--403.html.twig

It's also not ideal because someone could change that url for user.login in route and then that path would no longer work.

So I'm for removing it but decided not to get too opinionated with my changes.

chriswales95:
> If a site has set custom 404 and 403 pages in /admin/config/system/site-information would this override them? I wouldn't want to stop our [404 page](https://www.brighton-hove.gov.uk/404) working.

Hi @andybroomfield, No. This just overrides the default error messages. The user can still set their own defaults. What this does is put the previously hard-coded messages from localgov_base into localgov_core. The messages are, at the moment, hard-coded in the template. It doesn't make sense to be there so I think it makes more sense being in localgov_core if the slightly bespoke messages are still wanted.

If a user wants a custom error page, the redirect to the custom pages happens in the CustomPageExceptionHtmlSubscriber event subscriber which I have not touched. That's a core subscriber which checks if the user has set any custom pages and then redirects them.

If you look at the issue I mentioned above in the description it all stems from issues with the custom error pages which are a bit broken at the moment in terms of custom content.

There's some more info in the localgov_base issue and PR which I've linked below:
https://github.com/localgovdrupal/localgov_base/issues/535
https://github.com/localgovdrupal/localgov_base/pull/544

markconroy:
Hi @chriswales95 

I have merged the issue https://github.com/localgovdrupal/localgov_base/pull/544 so that we don't have hardcoded status messages in the 403/404 pages any more. 

With that being the case, do we need this PR? As in, maybe we should just use what core gives us, and let individual councils override that if they wish?

===

chriswales95:
Hi @markconroy, I'm happy for you to close. 

The reason I opened this was to preserve the customised messages that was in the templates. I wasn't sure if anyone was particularly attached to the messaging or if there was a usability need behind it (e.g. the login URL). 

It doesn't affect my needs so happy for you to close :).

Thanks,
Chris

finnlewis:
Thanks for this @chriswales95 @markconroy and @andybroomfield 

Sounds like we can close this for now.

PR #221, opened 2024-05-23
Move 'Files' tab to position it under admin/content/media

### LINKED ISSUE #220: Hide the 'files' tab to non admins
The files tab is pretty much meaningless to content designers, they need the 'media' tab. 

Let's hide it for all, except admin users.

Image

markconroy:
Closes #220 

Removes the permission to access the files overview permission. It's probably only developers that need this for debugging things, so we can add that manually when needed, or add it to the admin role (when we have one).

===


modules/localgov_roles/localgov_roles.install
```diff
@@ -13,4 +15,28 @@ function localgov_roles_install() {
...
+ */
+function _remove_access_files_overview_permission() {
+  $roles = Role::loadMultiple();
+  foreach ($roles as $role) {
+    if ($role->hasPermission('access files overview')) {
+      $role->revokePermission('access files overview');
+      $role->save();
+    }
+  }
+}
```
andybroomfield: I'm not so sure of this, if there are custom roles that have been assigned this then they will get deleted. Sometimes it is useful if there are files which for what ever reason are not yet in media. I'd limit this to localgov drupal provided roles, and I'd check if they have been altered from default. Noting the update hook also runs this.

stephen-cox:
As @andybroomfield points out - there are potential issues removing this permission for existing sites. Maybe just remove the update hook. We could always add a release note suggesting existing sites do this manually if it's important.

finnlewis:
Just discussing in Merge Tuesday.

Generally happy to remove the permission for fresh installs.

However, deploying an update hook that removes permissions from existing roles is perhaps a bit too far.

Perhaps we should remove the permission in the default install, and add instructions for people to clarify that they can remove the permission for certain roles.

finnlewis:
Note: would like to check where the permission 'access files overview' is being set to see if we can remove it there rather than here.

stephen-cox (commented):
There's been some discussion on #220 on how desirable this actually is. At least one person finds the files list useful, but it's also needed to ensure a file has been deleted.

If we go ahead and merge this we need to at least document how to get it back should it be needed.

markconroy:
Maybe we could have the best of both worlds.

1. Remove the files menu link from the admin/content page
2. Add the files menu link to the tabs on the admin/content/media page

That would mean people can still get to the files tab, but don't mistake if for the media tab which is what they usually want. It might make more sense to go to "Admin > Content > Media > Files" for content editors, as then it looks like files are a sub-item of media items (which they are).

ekes:
MT: moving the link sounds more promising than removing permission; 
maybe check with @willguv or product group about moving it being good.

Are there files being used that aren't in media?

finnlewis:
Note: @willguv is away for a few weeks, so might want to park this until he's back.

finnlewis:
@willguv likes the sound of @markconroy's suggestion!

markconroy:
I've rewritten this PR. New version:

1. Adds the `views.view.files.yml` view to our `config/install` so new sites can use it.
2. Adds a local task for the Files tab to sit under 'Media' on the `/admin/content/*` pages
3. Adds an update hook to update current sites to move 'Files' from admin/content/files to admin/content/media/files

---

localgov_core.install
stephen-cox, line 1: I don't think we should be moving the files tab for existing installs. The new location isn't very obvious and will be confusing to people who are using files already.

modules/localgov_media/config/install/views.view.files.yml
stephen-cox, line 1: This should be in the optional directory rather than install.

stephen-cox (changes requested):
I have tested this and it works, I do have a couple of comments though.

localgov_core.install
markconroy: I've removed the update hook now, however given that we are creating a `localgov_core.links.task.yml` file, we will now have a 'Files' tab on the admin/content page and also on the `admin/content/media` page. 

Maybe that's okay, and existing sites can update the view manually if they want to remove the tab from the `admin/content` page.

stephen-cox: There is a little oddity here for existing installs as going to the files listing now looks like this 

image

Let's discuss as a Merge Tuesday as to how much if an issue it is.

I also note that Files appears under Content in the menu and where perhaps it should under Content > Media. But if this is a problem then it could be another issue.

finnlewis:
@willguv we're discussing this in Merge Tuesday and not sure of the motivation for moving the files tab.

Could we discuss with content and product people? 

Files and media entities are different things, and there are use cases for accessing both. 

We could perhaps address this with a how-to or other documentation?

markconroy:
@finnlewis Will stated at product group meeting today that we should just create an update hook to remove the tab from admin/content and write a release note to say where the files tab has been moved to.

https://github.com/localgovdrupal/localgov_core/issues/220#issuecomment-2374125908

ekes:
referencing #283 follow-up for why this is done, and potentially reviewing how this is done.

PR #229, opened 2024-06-14
feat: use shared workflows

millnut:


finnlewis (approved):
Looking good, tests running and passing, thanks @millnut !

PR #230, opened 2024-06-17
fix: additional strict type fixes

millnut:
Fixes the following errors, reported in https://github.com/localgovdrupal/localgov_project/pull/153

```
 ------ ------------------------------------------------------------ 
  Line   modules/contrib/localgov_core/localgov_core.module          
 ------ ------------------------------------------------------------ 
  28     Only booleans are allowed in an if condition, mixed given.  
  31     Only booleans are allowed in an if condition, mixed given.  
 ------ ------------------------------------------------------------ 

 ------ -------------------------------------------------------------------- 
  Line   modules/contrib/localgov_core/src/Plugin/Block/PageHeaderBlock.php  
 ------ -------------------------------------------------------------------- 
  130    Only booleans are allowed in &&, mixed given on the right side.     
 ------ -------------------------------------------------------------------- 
```

### DIFF
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -25,10 +25,13 @@ function localgov_core_theme($existing, $type, $theme, $path) {
  * Implements hook_template_preprocess_default_variables_alter().
  */
 function localgov_core_template_preprocess_default_variables_alter(&$variables) {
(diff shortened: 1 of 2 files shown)


markconroy (approved):
This looks good to me.

===

PR #231, opened 2024-07-03
adds a dedicated card view mode

### LINKED ISSUE #57: Add view mode for 'Card'
Lots of websites use a "Card" view mode. Our current profile only allows teaser as a view mode, which is fine for some listings like a search result or service listing, but often we might like a slightly different layout, such as a featured news like what is on the Brighton-Hove homepage.

Screenshot 2021-04-09 at 15 11 57

Once we have this as a view mode, I'd suggest we place a template called `node--card.html.twig` in the `localgov_skeleton` theme so all themes can use the HTML from it and not invent their own per theme/site.

markconroy:
Closes #57 

---

### DIFF
--- a/config/install/core.entity_view_mode.node.localgov_card.yml
+++ b/config/install/core.entity_view_mode.node.localgov_card.yml
@@ -1,0 +1,10 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: node.localgov_card
+label: 'LocalGov Card'
+description: ''
+targetEntityType: node
+cache: true
--- a/localgov_core.install
+++ b/localgov_core.install
@@ -57,3 +57,19 @@ function localgov_core_update_8002() {
     }
   }
 }
+
+/**
+ * Add a localgov_card view mode.
+ */
+function localgov_core_update_8003() {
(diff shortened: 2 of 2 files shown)


ekes:
Woo... popped in here to see if there was anything backend-ish to check and there ain't so much. It adds to existing sites, but uses the `localgov_` name space so that should be safe. So pass on that, but while I'm already here, questions:

I know card display modes are pretty common, is the plan to use it somewhere in the default localgov install? I'd assume that would be the reason to want to add it to core? At the moment it's going to have all the default fields because there's no settings for content types? Is there a plan for where this is going?

andybroomfield:
Does this actully enable it on anything though? Is it planned to be used anywhere?

markconroy:
Yes, we should start using it for the featured news at the top of news listings for example.

We could eventually start shipping teaser and card view modes for each of our content types (that can be then used in views or in entity reference fields).

I already have a follow-up issue created in [LocalGov Base](https://github.com/localgovdrupal/localgov_base/issues/568) to add a card template.

Based on the work for elections recently, I can also imagine a "timeline" view mode in our not-too-distant future.

markconroy:
@andybroomfield @ekes are you okay for us to mark this as 'approved'?

@msayoung any thoughts?

andybroomfield (approved):

PR #233, opened 2024-07-12
Remove invalid attributes from the WYSIWYG editor's allowed tags

### LINKED ISSUE #232: Error on WYSIWYG text format config page from LinkIt
The following error message is displayed on /admin/config/content/formats/manage/wysiwyg when trying to save the settings.

The following attribute(s) are already supported by enabled plugins and should not be added to the Source Editing "Manually editable HTML tags" field: Linkit (\<a data-entity-type data-entity-uuid data-entity-substitution\>).

Looks to come from the config for this editor coming from the LocalGov Media module

https://github.com/localgovdrupal/localgov_core/blob/dc631ad7e9ade140d08ae01270dc380b5afa1c09/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml#L102

stephen-cox:
Fixes #232

### DIFF
--- a/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
@@ -99,7 +99,7 @@ settings:
         - '<h4 id>'
         - '<h5 id>'
         - '<h6 id>'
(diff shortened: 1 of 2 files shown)


ekes (approved):
Manually tested both ways. Works.

PR #234, opened 2024-07-16
adds admin role module

markconroy:
Adds a module to set an admin role.

---

### DIFF
--- a/modules/localgov_admin_role/localgov_admin_role.install
+++ b/modules/localgov_admin_role/localgov_admin_role.install
@@ -5,6 +5,7 @@
  * Install, update and uninstall functions for the localgov_admin_role module.
  */
 
+use Drupal\localgov_roles\RolesHelper;
 use Drupal\user\Entity\Role;
 
 /**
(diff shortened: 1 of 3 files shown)


modules/localgov_admin_role/localgov_admin_role.install
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+use Drupal\user\Entity\Role;
+
+/**
+ * Implements hook_install().
+ */
+function localgov_admin_role_install() {
+  // Check if the 'localgov_admin_role' role exists.
+  $admin_role = Role::load('localgov_admin');
```
stephen-cox: I suggest replacing the string by a class constant in the RolesHelper class, as this is what we're doing with other roles and makes the constant available should we wish to do any manipulation of this role later on.

https://github.com/localgovdrupal/localgov_core/blob/1e3c35bc65b1c97fae390cdaeccf45a0eb6f74fc/modules/localgov_roles/src/RolesHelper.php#L8-L28

stephen-cox (commented):
I think this approach is fine. Just one suggestion on using a class constant for the role name rather than a string.

markconroy:
Follow up issue to enable this module by default: https://github.com/localgovdrupal/localgov/pull/749

PR #239, opened 2024-08-10
fix: replace Drupal ContainerAwareEventDispatcher with Symfony EventDispatcher

millnut:
## What does this change?

Replaces Drupal ContainerAwareEventDispatcher with Symfony EventDispatcher.

ContainerAwareEventDispatcher was deprecated in Drupal 10.3, removed in Drupal 11 (see https://www.drupal.org/node/3376090)


finnlewis:
@millnut is this still relevant?

millnut:
Let me re-check this @finnlewis and get back to you

Issue #240, opened 2024-08-15
Enabling a module with config-import fails when setting permissions that don't exist yet


Polynya:
Installing LocalGov Publications using config import fails because `localgov_roles_modules_installed()` tries to add permissions before they exist - the new content type hasn't been imported yet. This probably applies to other LGD modules that create new content types.

I suggest  `localgov_roles_modules_installed()` does nothing if $is_syncing is true. Can we assume that all permissions have been set already in config?

Steps to reproduce with DDEV:
```
ddev drush si localgov -y
ddev export-db --file=lgd.sql.gz
ddev drush en localgov_publications
ddev drush cex
ddev import-db --file=lgd.sql.gz
ddev drush deploy
```
```

Error Output:                                                                                                 
================                                                                                              
 [notice] Synchronized extensions: install book.                                                              
 [notice] Synchronized extensions: install layout_discovery.                                                  
 [notice] Synchronized extensions: install layout_paragraphs.                                                 
 [notice] Synchronized extensions: install localgov_paragraphs_layout.                                        

In Role.php line 210:                                                                                         

Adding non-existent permissions to a role is not allowed. The incorrect permissions are "create localgov_p  
ublication_page content", "create localgov_publication_cover_page content", "delete any localgov_publication  
_page content", "delete any localgov_publication_cover_page content", "delete localgov_publication_page revi  
sions", "delete localgov_publication_cover_page revisions", "delete own localgov_publication_page content",   
"delete own localgov_publication_cover_page content", "edit any localgov_publication_page content", "edit an  
y localgov_publication_cover_page content", "edit own localgov_publication_page content", "edit own localgov  
_publication_cover_page content", "revert localgov_publication_page revisions", "revert localgov_publication  
_cover_page revisions", "view localgov_publication_page revisions", "view localgov_publication_cover_page re  
visions".                                                                                                     

```

PR #241, opened 2024-08-15
Do not call assignModuleRoles during config import.

Polynya:
See https://github.com/localgovdrupal/localgov_core/issues/240
 
Installing LocalGov Publications using config import fails because `localgov_roles_modules_installed()` tries to add permissions before they exist - the new content type hasn't been imported yet. This probably applies to other LGD modules that create new content types.

## What does this change?

Change localgov_roles_modules_installed() to do nothing if $is_syncing is true.

## How to test

```
ddev drush si localgov -y
ddev export-db --file=lgd.sql.gz
ddev drush en localgov_publications
ddev drush cex
ddev import-db --file=lgd.sql.gz
ddev drush deploy
```

No errors running `drush deploy`. 

The workaround is to run `drush config-import` again but isn't great when running pipelines. Also, is there a risk that the error stops other processes running which won't be attempted on the next config import?

 Can we assume that all permissions have been set already in config?

### DIFF
--- a/modules/localgov_roles/localgov_roles.module
+++ b/modules/localgov_roles/localgov_roles.module
@@ -10,8 +10,10 @@ use Drupal\localgov_roles\RolesHelper;
 /**
  * Implements hook_modules_installed().
  */
-function localgov_roles_modules_installed($modules) {
-  foreach ($modules as $module) {
-    RolesHelper::assignModuleRoles($module);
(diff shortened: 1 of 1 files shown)


finnlewis:
Makes sense to @ekes  and @stephen-cox ! :)

Thanks @Polynya !

Issue #245, opened 2024-08-23
Installing site from existing config fails


stephen-cox:
Error is: 'block' entity with ID 'localgov_base_publication_navigation' already exists.

## Steps to reproduce

Delete the install_localgov hook from the profile or patch core from https://www.drupal.org/project/drupal/issues/2982052

```
drush si localgov -y
drush localgov_workflows localgov_publications -y
drush cex -y
drush si localgov --existing-config -y
```

## Exception trace
```
  at /app/web/core/lib/Drupal/Core/Entity/EntityStorageBase.php:519
 Drupal\Core\Entity\EntityStorageBase->doPreSave() at /app/web/core/lib/Drupal/Core/Entity/EntityStorageBase.php:483
 Drupal\Core\Entity\EntityStorageBase->save() at /app/web/core/lib/Drupal/Core/Config/Entity/ConfigEntityStorage.php:257
 Drupal\Core\Config\Entity\ConfigEntityStorage->save() at /app/web/core/lib/Drupal/Core/Entity/EntityBase.php:354
 Drupal\Core\Entity\EntityBase->save() at /app/web/core/lib/Drupal/Core/Config/Entity/ConfigEntityBase.php:614
 Drupal\Core\Config\Entity\ConfigEntityBase->save() at /app/web/modules/contrib/localgov_core/src/Service/DefaultBlockInstaller.php:136
 Drupal\localgov_core\Service\DefaultBlockInstaller->installForModule() at /app/web/modules/contrib/localgov_core/src/Service/DefaultBlockInstaller.php:109
 Drupal\localgov_core\Service\DefaultBlockInstaller->install() at /app/web/modules/contrib/localgov_core/localgov_core.module:73
 localgov_core_localgov_post_install() at /app/web/profiles/contrib/localgov/localgov.profile:45
 {closure}() at /app/web/core/lib/Drupal/Core/Extension/ModuleHandler.php:395
 Drupal\Core\Extension\ModuleHandler->invokeAllWith() at /app/web/profiles/contrib/localgov/localgov.profile:46
 localgov_post_install_task() at /app/web/core/includes/install.core.inc:697
 install_run_task() at /app/web/core/includes/install.core.inc:574
 install_run_tasks() at /app/web/core/includes/install.core.inc:122
 install_drupal() at /app/vendor/drush/drush/includes/drush.inc:69
 drush_call_user_func_array() at /app/vendor/drush/drush/includes/drush.inc:53
[...]

PR #249, opened 2024-09-16
Fix/2.x/245 installing site from existing config fails

rupertj:
Checks if a block with the derived ID of a new block exists before trying to save a new block.

Adds and checks the $is_syncing param of hook_modules_installed().

Fix for #245


stephen-cox (approved):
This does work on the 1.x branch of Publications, but not the 1.0.6 release. I've created a PR for a 1.0.7 release https://github.com/localgovdrupal/localgov_publications/pull/204

PR #257, opened 2024-11-13
Add a view display extender for the localgov_page_header for the lede

### LINKED ISSUE #127: How to add a lede (subtitle) to a view?
Has anyone worked this out, would a custom views plugin be needed?
This is for the page header block.

andybroomfield:
Fix #127

## What does this change?

Adds a setting to views that allows for a page header category, with the option
to set a custom lede inside the view, which will then be used by the
page header block.

## How to test
1. Enable the new page header display extender on /admin/structure/views/settings/advanced (should we enable this by default?)
2. Create a new view with a page display.
3. There should be a new page header section in the center of the views configuration.
4. Edit that and add some summary text, you should also be able to use tokens from the view.
5. View the views page, and the page header should also display the entered summary as the lede (part of the page header block).

Now possible for site builders to add subtitles to views.

May conflict with others who have added their own extension to get a subtitle on the view.

## Images

Screenshot 2024-11-13 at 9 47 10 am

Screenshot 2024-11-13 at 9 52 19 am

Screenshot 2024-11-13 at 9 52 23 am

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/src/Plugin/Block/PageHeaderBlock.php
+++ b/src/Plugin/Block/PageHeaderBlock.php
@@ -14,6 +14,8 @@ use Drupal\Core\Session\AccountInterface;
 use Drupal\localgov_core\Event\PageHeaderDisplayEvent;
 use Drupal\node\Entity\Node;
 use Drupal\taxonomy\Entity\Term;
+use Drupal\views\ViewExecutable;
+use Drupal\views\Views;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
 
@@ -130,6 +132,7 @@ class PageHeaderBlock extends BlockBase implements ContainerFactoryPluginInterfa
     $route = $this->currentRouteMatch->getRouteObject();
     if (!is_null($route)) {
       $parameters = $route->getOption('parameters');
+      $defaults = $route->getDefaults();
       if (!is_null($parameters)) {
         foreach ($parameters as $name => $options) {
           if (!isset($options['type'])) {
@@ -150,6 +153,13 @@ class PageHeaderBlock extends BlockBase implements ContainerFactoryPluginInterfa
             $this->entity = $entity;
             break;
           }
+        }
+      }
+      if (isset($defaults['view_id']) && isset($defaults['display_id'])) {
+        $view = Views::getView($defaults['view_id']);
+        if ($view) {
+          $view->setDisplay($defaults['display_id']);
+          $this->entity = $view;
         }
       }
     }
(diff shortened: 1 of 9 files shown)


src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -152,6 +155,13 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
           }
         }
       }
+      if (isset($defaults['view_id']) && isset($defaults['display_id'])) {
+        $view = Views::getView($defaults['view_id']);
+        if ($view) {
+          $view->setDisplay($defaults['display_id']);
+          $this->entity = $view;
```
andybroomfield: Not of type Drupal\Core\Entity\EntityInterface. 
The pageHeaderEvent isn't doing type checking (maybe it should) so perhaps we need to send $view separately?

finnlewis:
@Adnan-cds can see this will be potentially useful!

Adnan-cds:
It works!  Thanks Andy :)

The code looks okay at first glance.  I would suggest that Steve @stephen-cox does a thorough review.

finnlewis:
Some questions from Merge Tuesday: 

1. Do we want to enable this by default?
2. How do we enable this by default? (Follow how MetaTag does this?)
3. Code review: happy with the approach?

src/Plugin/views/display_extender/PageHeaderDisplayExtender.php
```diff
@@ -0,0 +1,251 @@
...
+   *
+   * @var array
+   */
+  protected static $firstRowTokens;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    /** @var \Drupal\metatag_views\Plugin\views\display_extender\MetatagDisplayExtender */
```
stephen-cox: This type hint references the Metatag display extender and isn't needed here anyway

stephen-cox (commented):
I'm happy with this approach. One small nitpick with a type hint that's wrong and can probably be removed.

This does add a hard dependency between Core and Views. I don't have a problem with this as I can't think of an LGD site that wouldn't use Views, but we should add the dependency in the info.yml file.

I think it would also be good to enable this by default and maybe add an update hook to enable it for others. The test already includes the code on how to do this.

https://github.com/localgovdrupal/localgov_core/blob/c9c7ad9cafa1096e425c08e31c1fc38d3cb8560a/tests/src/Functional/ViewsPageExtenderTest.php#L40-L44

stephen-cox:
@stephen-cox Will test the update hook and merge if okay

andybroomfield:
Discussed on MT on 28/1/2025: We'd like to get this reviewed soon. @ekes @stephen-cox please.

stephen-cox (approved):
Looking good to me :+1:

PR #258, opened 2024-11-24
fix: add return types

millnut:
## What does this change?

Added some minor rector fixes in preparation for https://www.drupal.org/node/3461320

### DIFF
--- a/tests/src/Functional/OnOffSwitchTest.php
+++ b/tests/src/Functional/OnOffSwitchTest.php
@@ -24,7 +24,7 @@ class OnOffSwitchTest extends BrowserTestBase {
   /**
    * Data provider for testOnOffSwitch().
    */
-  public static function onOffProvider() {
+  public static function onOffProvider(): \Generator {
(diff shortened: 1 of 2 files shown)


tests/src/Functional/OnOffSwitchTest.php
```diff
@@ -24,7 +24,7 @@ class OnOffSwitchTest extends BrowserTestBase {
   /**
    * Data provider for testOnOffSwitch().
    */
-  public function onOffProvider() {
+  public static function onOffProvider() {
```
ekes: Nit pick. If we're adding return types this should be `: \Generator` ?

ekes (changes requested):

tests/src/Functional/OnOffSwitchTest.php
millnut: Good flag. Not sure why rector didn't flag this but I've updated the PR

ekes (approved):

PR #260, opened 2024-11-28
adds language widget to ckeditor

### LINKED ISSUE #259: Add the Language widget to the default Text editor profile
In order to meet the WCAG criteria 3.1.2 : Language of Parts authors need to be able to identify phrases of content that are written in different (human) language from the surrounding text. 

This allows assistive technology to process it properly - for example a screen reader will attempt to pronounce it correctly. 

An example on our demo site: https://demo.localgovdrupal.org/national-parks-uk/pembrokeshire-coast . The Welsh name:  Parc Cenedlaethol Arfordir Penfro could be wrapped in a `<span lang="cy-GB">` 

(Note: proper names are an exception on the criteria so don't have to conform, but they should)

Recommendation: 
Add the language switcher to the Text editor by default. I suggest we allow Drupal's 107 predefined languages.

markconroy:
Closes #259

## What does this change?

Adds language widget to ckEditor
Sets it to allow languages enabled for site (currently only English on a default install)

## How to test

Add some content via wysiwyg. Select a word or two, choose "Language" from the ckEditor toolbar, and set a language for that selection. The resultant markup for it should then be something like `<span dir="ltr" lang="en">selected text</span>`

---

### DIFF
--- a/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
@@ -29,9 +29,10 @@ settings:
       - drupalInsertImage
       - drupalMedia
       - '|'
-      - sourceEditing
-      - '|'
       - insertTable
+      - '|'
+      - sourceEditing
+      - textPartLanguage
   plugins:
     ckeditor5_heading:
       enabled_headings:
@@ -41,6 +42,34 @@ settings:
         - heading4
         - heading5
         - heading6
+    ckeditor5_imageResize:
+      allow_resize: true
+    ckeditor5_language:
+      language_list: site_configured
+    ckeditor5_list:
+      properties:
+        reversed: false
+        startIndex: true
+      multiBlock: true
+    ckeditor5_sourceEditing:
+      allowed_tags:
+        - '<cite>'
+        - '<dl>'
+        - '<dt>'
+        - '<dd>'
+        - '<mark>'
+        - '<blockquote cite>'
+        - '<ol type>'
+        - '<h2 id>'
+        - '<h3 id>'
+        - '<h4 id>'
+        - '<h5 id>'
+        - '<h6 id>'
+        - '<a hreflang title>'
+        - '<ul type>'
+        - '<img data-entity-type data-entity-uuid data-caption>'
+        - '<drupal-media data-caption title>'
+        - '<span>'
     ckeditor5_style:
       styles:
         -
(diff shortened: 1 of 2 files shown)


modules/localgov_media/config/optional/filter.format.wysiwyg.yml
```diff
@@ -57,21 +44,34 @@ filters:
...
+      allowed_html: '<br> <p class="alert alert-info alert-danger alert-fail alert-success callout callout-primary callout-success callout-danger callout-teal callout-carbon callout-yellow"> <h1> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <a class="external-link pdf-link btn btn-start" hreflang title href data-entity-type data-entity-uuid data-entity-substitution> <ul class="list-checked" type> <cite> <dl> <dt> <dd> <mark> <blockquote cite> <ol type start> <img data-entity-type data-entity-uuid data-caption src alt height width data-align> <drupal-media data-caption title data-entity-type data-entity-uuid alt data-view-mode data-align> <span dir> <strong> <em> <u> <li> <table> <tr> <td rowspan colspan> <th rowspan colspan> <thead> <tbody> <tfoot> <caption>'
```
msayoung: I don't see `<code>` `<pre>` in the list here.

modules/localgov_media/config/optional/editor.editor.wysiwyg.yml
```diff
@@ -85,39 +114,10 @@ settings:
...
-      allow_view_mode_override: true
     linkit_extension:
       linkit_enabled: true
       linkit_profile: default
+    media_media:
+      allow_view_mode_override: true
 image_upload:
   status: false
-  scheme: public
-  directory: inline-images
```
msayoung, line 119: this isn't in the new config. not sure if its needed?

markconroy: I didn't do anything to the image ckeditor widget. But I reckon those lines are in there as a hangover from when we used to allow images to be uploaded via wysiwyg, so it would need a `scheme: public` and `directory: inline-images`. For ages now, we have only been allowing external images via the wysiwyg which is why the line above is set to `false`. I have a feeling anytime anyone updates the  "Text Formats and Filters" in their sites and does a `config export` they'll get what we have here.

modules/localgov_media/config/optional/filter.format.wysiwyg.yml
markconroy: They are back in now.

msayoung:
Great - now I'm wondering which languages we should have as options. 

The options are: 

The list of languages in the CKEditor "Language" dropdown can present the [6 official languages of the UN](https://www.un.org/en/sections/about-un/official-languages), all 107 languages predefined in Drupal, or the languages configured for this site.

We're currently showing the languages configured for the site (English)

I'm not sure if we want to change this?

PR #266, opened 2024-12-31
style(standards): fix eslint js errors

millnut:
## What does this change?

Updates the JS to match Drupal Core eslint js standards

~Do not merge yet as eslint is not seeing the core prettier config so thinks `'` need replacing with `"` which is incorrect as the core prettier config has `"singleQuote": true,`~ see https://github.com/localgovdrupal/localgov_shared_workflows/pull/21 on failure

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   tests:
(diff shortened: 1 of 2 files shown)


ekes (approved):
I'm going to approve this. It's almost all white space changes (except `function` to { `() => {`). 
But I'll refrain from directly merging (I've merged some very simple ones) so a second pair of eyes go on it first.

PR #267, opened 2025-01-16
disables html5 validation

markconroy:
Copy of [](https://github.com/localgovdrupal/localgov/pull/800) to use here instead of there.

### DIFF
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -5,6 +5,7 @@
  * LocalGovDrupal Core module file.
  */
 
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Installer\InstallerKernel;
 
 /**
@@ -76,4 +77,11 @@ function localgov_core_localgov_post_install(): void {
   /** @var \Drupal\localgov_core\Service\DefaultBlockInstaller $defaultBlockInstaller */
   $defaultBlockInstaller = \Drupal::service('localgov_core.default_block_installer');
   $defaultBlockInstaller->install($localgovModules);
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function localgov_core_form_alter(&$form, FormStateInterface $form_state, $form_id): void {
+  $form['#attributes']['novalidate'] = 'novalidate';
 }


ekes:
Immediate thought: To install most localgov_ modules you have to have this module. This disabled HTML5 validation on all forms. Plenty of sites have been extend to have non-localgov_ modules installed, could any of these modules, or things otherwise built into the site, be expecting HTML5 validation to be enabled?

andybroomfield:
Suggest we add a setting
```php
$settings['localgov_enable_html5_validation'] = TRUE;
```
and detect it using
```php
if (Settings::get('localgov_enable_html5_validation') !== TRUE) {
... remove html5 validation.
```
This is because as above there may be cases where modules would expect html5 validation to be present, or may not have any other validation. So whilst we should encourage the GDS recommendation of using validation errors that allow  for more descriptive text, if there are cases where this isn't set up correct there is a chance to opt out of this change.

finnlewis:
Or we could just use the module which allows people to turn it off easily. 

https://www.drupal.org/project/disable_html5_validation

Then we enable it on a default install, but it is optional for existing installs.

Thoughts @andybroomfield @ekes @markconroy ?

finnlewis:
Discussing in Merge Tuesday, general agreement that it would be good to make it optional.

@andybroomfield mentioned that in general it would be good to reduce dependencies. 

I'm in favour of using the module! 

hmmm.... to be discussed further....

markconroy:
I have no preference for doing this via the contrib module or via a custom hook. 

It's surprising that Drupal core has such good form validation, but doesn't disable it by default. So maybe we should go with the module approach to follow that practice.

ekes:
I'm generally not in favour of adding modules for one line that you could include elsewhere. OK maybe more than generally. I can't think of another occasion I think it's a good idea. But in this case. It feels like it makes sense. It is explicit. It's easy for end UI admins to understand.

andybroomfield:
Discussed on MT on 28/1/2025. We thought that switching off the module is better as it allows site administrators to turn it off if they encountered a problem. We will install the module in the profile and then switch it on by default.

finnlewis:
Allrighty then... closing this in favour of https://github.com/localgovdrupal/localgov/pull/820

PR #271, opened 2025-02-14
Fix NULL $extender breaks PageHeaderBlock

### LINKED ISSUE #270: NULL $extender breaks PageHeaderBlock
Fault at line: https://github.com/localgovdrupal/localgov_core/blame/ced43a28fc9f52d5951a66fb282ea191fc1c007c/src/Plugin/Block/PageHeaderBlock.php#L261

Issue: If $extender is NULL, the lede part of the code breaks with:
`
Error: Call to a member function getLede() on null in Drupal\localgov_core\Plugin\Block\PageHeaderBlock->getLede() (line 265 of modules/contrib/localgov_core/src/Plugin/Block/PageHeaderBlock.php).
`

Solution: Provide check for null on the extender variable

andybroomfield:
Fix #270.

Since the PageHeaderDisplay extender can be disabled, support cases when it is not enabled.

## What does this change?

- Amend viewsPageExtenderTest to account for page header extender being disabled
- Fix pageHeaderBlock to account for when views page header display extender is disabled

## How to test

Under /admin/structure/views/settings/advanced turn off Page header display extender.
Then load a views page (or create a new page view).
Should no longer WSOD.

Less WSOD reports

Fallback is to NULL so restores previous behaviour.

## Images

n/a

## Accessibility

n/a

### DIFF
--- a/src/Plugin/Block/PageHeaderBlock.php
+++ b/src/Plugin/Block/PageHeaderBlock.php
@@ -18,6 +18,7 @@ use Drupal\views\Views;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Drupal\localgov_core\Plugin\views\display_extender\PageHeaderDisplayExtender;
 
 /**
  * Provides a 'PageHeaderBlock' block.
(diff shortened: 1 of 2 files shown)


andybroomfield:
//cc @NikLP

ekes (approved):
As discussed in slack. Correct that it returns NULL if not existing.

PR #274, opened 2025-03-02
Remove patch for paragraphs module.

NikLP:
Latest composer update shows that this patch in composer.json for 3492419 no longer applies. 

Investigation shows the issue is fixed and released in https://www.drupal.org/project/paragraphs/releases/8.x-1.19 

Patch duly removed here. Not personally tested it, but here's the work.


finnlewis (approved):
Thanks @NikLP  - works for me :)

PR #276, opened 2025-03-04
feat: allow image_widget_crop 3 for D11 compatibility

millnut:
## What does this change?

This allows image_widget_crop 3 so we can have D11 compatibility as well as the existing version 2

Version 3
```
Works with Drupal: ^9.5 || ^10 || ^11
```

Version 2
```
Works with Drupal: ^8 || ^9 || ^10
```

Users can choose to fix to version 2 if they are on D10, however we can discuss next MT if we want to go straight to version 3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "minimum-stability": "dev",
     "require": {
         "drupal/field_group": "^3.1",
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Works for me, thanks @millnut !!!

PR #278, opened 2025-03-14
Add character counting

stephen-cox:
## What does this change?

Adds a UI to easily enable character counting on test fields.

## How to test

Checkout this branch and enable the localgov_char_count module. The settings form is at /admin/config/content/localgov-char-count which should allow you to turn on and off character counting for title and summary fields for LGD content types.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
         "drupal/metatag": "^2.0.2",
         "drupal/pathauto": "^1.8",
         "drupal/role_delegation": "^1.1",
+        "drupal/textfield_counter": "^2.4",
         "drupal/token": "^1.7"
     },
     "require-dev": {
--- a/modules/localgov_char_count/README.md
+++ b/modules/localgov_char_count/README.md
@@ -1,0 +1,1 @@
+##  LocalGov Character Count module
--- a/modules/localgov_char_count/config/install/localgov_char_count.settings.yml
+++ b/modules/localgov_char_count/config/install/localgov_char_count.settings.yml
@@ -1,3 +1,3 @@
 title_length: 60
 summary_length: 160
-status_message: '@current_length / @maxlength characters'
+status_message: '<span class="current_count">@current_length</span> / <span class="maxlength_count">@maxlength</span> characters'
--- a/modules/localgov_char_count/config/install/localgov_char_count.settings.yml
+++ b/modules/localgov_char_count/config/install/localgov_char_count.settings.yml
@@ -1,0 +1,3 @@
+title_length: 60
+summary_length: 160
+status_message: '@current_length / @maxlength characters'
--- a/modules/localgov_char_count/config/schema/localgov_char_count.schema.yml
+++ b/modules/localgov_char_count/config/schema/localgov_char_count.schema.yml
@@ -1,5 +1,5 @@
 localgov_char_count.settings:
-  type: mapping
+  type: config_object
   label: 'LocalGov character counting settings'
   mapping:
     title_length:
--- a/modules/localgov_char_count/config/schema/localgov_char_count.schema.yml
+++ b/modules/localgov_char_count/config/schema/localgov_char_count.schema.yml
@@ -1,0 +1,13 @@
+localgov_char_count.settings:
+  type: mapping
+  label: 'LocalGov character counting settings'
+  mapping:
+    title_length:
+      type: integer
+      label: 'The maximum number of characters allowed in the title field.'
+    summary_length:
+      type: integer
+      label: 'The maximum number of characters allowed in the title field.'
+    status_message:
+      type: string
+      label: 'The message shown indicating the status of the character count. The variables @maxlength, @current_length, @remaining_count'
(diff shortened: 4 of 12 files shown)


finnlewis:
Discussing in Merge Tuesday - @stephen-cox asks should this live here in localgov_core or separately. 

@andybroomfield keen to avoid putting more optional code / dependencies into localgov_core

If this is a separate project, we should probably put it directly on drupal.org 

Discuss! 

CC @ekes @willguv @tonypaulbarker

stephen-cox:
I've added some tests so marking this as ready for review, but aware that it may be better to move this to it's own module on drupal.org. My only concern about that is this is only a very small module that is only specific to a very small use case and currently only applies to LocaGov Drupal content types.

stephen-cox:
@willguv I have deployed this to product site for you to test. The form for configuring character counting is here: https://product.localgovdrupal.org/admin/config/content/localgov-char-count

andybroomfield:
Follow up from Merge Tuesday discussion:

My preference would be for this to be a standalone utility module, rather than part of localgov_core. 

This relates to https://github.com/localgovdrupal/localgov/issues/625 which was referencing contrib modules, though I think the same principle applies. 

A core module that will get installed as a required dependency by most major tentpole modules from localgov drupal, this also means all the code will get installed, so it should cover generally required functionality to get things working. Previously product features where added to the profile as they where customising other modules to shape the product. They have since been spun out as their own utility modules ([localgov_login_redirect](https://github.com/localgovdrupal/localgov_login_redirect/releases/tag/1.0.0) and [localgov_login_redirect](https://github.com/localgovdrupal/localgov_content_lock/releases/tag/1.0.0)).

It's also quite possible that there may be incompatibilities in the utility modules that stops a major update to localgov_core, or forces a break in compatibility with a Drupal version or a dependency. This could be much more easily managed as seperate modules.

There's also the case that sometimes we need to pin to specific versions of a module if it breaks something on our site. With localgov_core this could mean missing out on important bug fixes due to an incompatibility in one of the sub modules. By having them as separate utility modules, the impact can be limited to just that module.

There are concerns about the maintainability of so many modules. I'm not so sure it would be the case as the work on reviewing and updating PRs should be roughly the same whether it was in core or a separate module. There would be extra work in making a release PR and tagging a release. Arguably though aside from major deprecations, past experience releases tend to happen quickly after a feature or fix PR is added, so it wouldn't be that different unless both core and the independent module were updated.

Separate modules fit with the principle of [Do One Thing and Do It Well](https://principles.design/examples/unix-philosophy#:~:text=Do%20One%20Thing%20and%20Do%20It%20Well).

stephen-cox:
I'm happy for this to be spun out into it's own module. Although I do wonder whether all these small utility modules would be easier to manage if they were in a `localgov_utility` module.

I've asked for review of this PR as it already existed with most of the work already done. This means people can review the changes here and if this gets spun out into a separate module this can be closed with the knowledge that the code has already been reviewed.

ekes:
So if I understand the purpose here is to make it easier to set the configuration on one page, rather than the site builder doing what the module does on each form component?

It does that. But only for one form mode (default). And for nodes. The assumption is that people who need it on other entity types, or other form modes will do the configuration themselves?

So with those assumptions seems good. Certainly feels like some sort of utility module to live somewhere.

willguv:
@stephen-cox @ekes I'd like this to be the start of a 'content quality' module 

I've talked about this at various meetings and events with content designers, and it lead to us including a 'content metrics' idea in the Community Fund https://localgovdrupal.org/products/community-fund

How about we start with this feature and add to it over time?

andybroomfield:
Have tested the sub module here and it works as described. Was a little confused as the settings form is really just setting this up on the manage form displays that textcounter uses. I'd be wary of allowing editor to alter this as this will change the node form display config which can be replaced during deployments.

stephen-cox:
In reply to @ekes 

> So if I understand the purpose here is to make it easier to set the configuration on one page, rather than the site builder doing what the module does on each form component?

Yep - the idea is it's easy to turn this on just for the titles and summaries of LGD content types.

> It does that. But only for one form mode (default). And for nodes. The assumption is that people who need it on other entity types, or other form modes will do the configuration themselves?

Yep - any changes beyond the default form mode for LGD content types needs to be done as a normal site building job.

finnlewis:
Discussing in Merge Tuesday. 

Sounds like the direction of travel is a localgov_utilities module on drupal.org and pop this there as a sub-module. 

@stephen-cox and @finnlewis will liaise on this.

willguv:
Did you talk about what I wrote earlier @finnlewis 

[@stephen-cox @ekes I'd like this to be the start of a 'content quality' module](https://github.com/localgovdrupal/localgov_core/pull/278#issuecomment-2756676165)

stephen-cox:
@willguv My concern about a content quality module is that what will go in it is still unclear; something like the content metrics is likely to be big enough to warrant it's own module.

A LGD utilities module could include little things that are LGD specific; such as the [LGD Content Lock](https://github.com/localgovdrupal/localgov_content_lock) and [Login Redirect](https://github.com/localgovdrupal/localgov_login_redirect) modules that were in the profile.

It is possible to move modules around later if it would help rationalise things.

stephen-cox:
There's a new [LocalGov Utilities](https://www.drupal.org/project/localgov_utilities) module and here's a PR adding this as a submodule https://git.drupalcode.org/project/localgov_utilities/-/merge_requests/1

Issue #194, opened 2023-11-18
Add the entity to the page header template and bundle template suggestions to page header block


andybroomfield:
> Coming over from the [PR 133 on `localgov_guides`](https://github.com/localgovdrupal/localgov_guides/pull/133). 
> 
> I like this (though I would quibble pointlessly about that subtitle markup :grinning:). I have two additional suggestions:
> 
> 1. add another variable, `node`, to the module's `hook_theme` implementation for this block, then populate it with the node object. This will make it simple, in themes, to override the rendering of the block according to content type (and according to the node content itself),
> 2. add template suggestions for the _content type_ to the block template (maybe using `EntityTypeBundleInfo()` in `hook_theme_suggestions_HOOK()`...not sure if we would want to limit this to LGD content types, or just catch them all),
> 
> The fact that what might be relatively simple things to handle in themes spawn issues like [consider stacked h1 pattern for guides](https://github.com/localgovdrupal/localgov_guides/issues/122), and that (as you pointed out) `localgov_publications` uses a _different_ block suggest to me that the core block could be more helpful specifically w/r/t the current content type.

_Originally posted by @ctorgalson in https://github.com/localgovdrupal/localgov_core/issues/193#issuecomment-1794389832_

As it can be used on any entity, perhaps entity type and bundle type.
The entity is already part of the block so can be added to the template.

rupertj:
I'd be well in favour of this. I'm pretty sure it'd let us get rid of the header block that localgov_publications has in favour of just using the existing one with a template and some preprocess. It'd also simplify some of the theme work we've done on H&F.

PR #279, opened 2025-04-03
Feature/2.x/194 add entity and bundle template suggestions to page header block

ctorgalson:
Re: https://github.com/localgovdrupal/localgov_core/issues/194

## What does this change?

This makes three non-breaking changes to `localgov_core`'s Page Header Block:

1. adds `#entity` to `localgov_page_header_block` in `localgov_core_theme()`
2. adds the current entity to `PageHeaderBlock.php`.s render array
3. adds Twig template suggestions to the block via `localgov_core_theme_suggestions_localgov_page_header_block()`

## How to test

There are five things requiring testing:

1. that _nothing changes_ on existing sites like https://demo.localgovdrupal.org
2. that three types of Twig template suggestions are provided:
  - per entity id, e.g. `localgov-page-header-block--node--7.html.twig`
  - per bundle, e.g. `localgov-page-header-block--localgov-services-landing.html.twig`
  - per entity type, e.g. `localgov-page-header-block--node.html.twig`
  
    image
3. that the expected entity is present in Twig; add the following to `localgov-page-header-block.html.twig` (or one of the suggested ones) and verify on the front end:
    ```twig
    {#
      On e.g. a localgov_services_landing node, with nid 7, this should produce the following output:
      
    0 array:3 [
      0 => "7"
      1 => "localgov_services_landing"
      2 => "node"
    ]
    
    #}
    {{ dump([
      entity.id(),
      entity.bundle(),
      entity.getEntityTypeId(),
    ]) }}
    ```
4. that templates named according to the suggestions are discovered and rendered
5. that non-entity pages (like Views) don't break because of these changes

Three ways:

1. if nothing whatever changes on any existing site
2. that it becomes possible to access the base entity from within the block template
3. that it becomes possible to override the module's template (and `localgov_base`'s template) with nid, bundle, and entity type specific templates.

As this is only intended to expand what's _possible_ without actually changing anything, it should be low risk (mainly the risk is that I've failed to grasp something important in the block plugin :smiley:)

## Images

n/a

## Accessibility

n/a

### DIFF
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -31,9 +31,12 @@ function localgov_core_theme_suggestions_localgov_page_header_block(array $varia
   $suggestions = [];
   $entity = $variables['entity'];
 
-  $suggestions[] = 'localgov_page_header_block__' . $entity->bundle();
-  $suggestions[] = 'localgov_page_header_block__' . $entity->id();
-  $suggestions[] = 'localgov_page_header_block__' . $entity->getEntityTypeId();
+  if ($entity instanceof NodeInterface || $entity instanceof TermInterface) {
+    $suggestions[] = 'localgov_page_header_block__' . $entity->bundle();
+    $suggestions[] = 'localgov_page_header_block__' . $entity->id();
+    $suggestions[] = 'localgov_page_header_block__' . $entity->getEntityTypeId();
+  }
+
   return $suggestions;
 }
 
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -17,6 +17,7 @@ function localgov_core_theme($existing, $type, $theme, $path) {
         'title' => '',
         'subtitle' => NULL,
         'lede' => NULL,
+        'entity' => NULL,
       ],
       'render element' => 'block',
     ],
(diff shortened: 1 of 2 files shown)


ctorgalson:
Looks like the still-failing test is not related to this change.

localgov_core.module
```diff
@@ -1,5 +1,8 @@
 <?php
 
+use Drupal\node\NodeInterface;
+use Drupal\taxonomy\TermInterface;
+
```
millnut: Hi @ctorgalson the use statements need to go below the file comment, could you add these line below the `use` statement on line 11 and see if that gets the tests to continue

millnut (changes requested):

localgov_core.module
ctorgalson: D'oh! Thanks for spotting that.

finnlewis:
Thanks @ctorgalson ! Let's try to test and review this week.

finnlewis:
@rupertj you might be interested in this one too!

markconroy (approved):
Works a treat, no impact on views pages.

PR #281, opened 2025-04-07
CSS/JS from localgov_base is not removed if it comes from an SDC component

### LINKED ISSUE #280: CSS/JS from localgov_base is not removed if it comes from an SDC component
Right now there are 2 components in localgov_base that have CSS and JS:
* add-to-calendar
* prev-next

When the sub-theme uses the `localgov_base_remove_css` or `localgov_base_remove_js` theme setting, those components' CSS and JS are not removed.

In regular templates in localgov_base, you have to manually add libraries to get the CSS/JS added to Drupal, so those templates simply check those settings before adding the libraries.

With components, no library definitions are needed. Drupal core's SDC will automatically create a library and auto add the library when the component is used.

That means we need a hook to check the theme settings and then remove the libraries if present.

JohnAlbin:
Fixes #280

## What does this change?

See #280 for the discussion.

## How to test

1. Add the `localgov_base:add-to-calendar.twig` component to an existing Drupal template. (For some reason I can't find any templates in LocalGov that are using the localgov_base components yet.)
2. Notice that the `add-to-calandar.css` and `add-to-calandar.js` files are added to the webpage when the component is used.
3. Turn on the "localgov_base_remove_css" and "localgov_base_remove_js" theme settings in your theme.
4. Notice that the `add-to-calandar.css` and `add-to-calandar.js` files are STILL added to the webpage when the component is used.

1. Apply the patch and retest with different combos of on/off for those theme settings. SDC creates a single library with both the CSS and the JS in it, but the new hook removes just the CSS or just the JS if that's what the theme settings are set to.

### DIFF
--- a/localgov_core.module
+++ b/localgov_core.module
@@ -24,6 +24,27 @@ function localgov_core_theme($existing, $type, $theme, $path) {
 }
 
 /**
+ * Implements hook_library_info_alter().
+ */
+function localgov_core_library_info_alter(&$libraries, $extension) {
+  if ($extension === 'core') {
+    $remove_css = (bool) theme_get_setting('localgov_base_remove_css');
+    $remove_js = (bool) theme_get_setting('localgov_base_remove_js');
+
+    foreach($libraries as $name => $library) {
+      if (str_starts_with($name, 'components.localgov_base--')) {
+        if ($remove_css) {
+          $libraries[$name]['css'] = [];
+        }
+        if ($remove_js) {
+          $libraries[$name]['js'] = [];
+        }
+      }
+    }
+  }
+}
+
+/**
  * Implements hook_template_preprocess_default_variables_alter().
  */
 function localgov_core_template_preprocess_default_variables_alter(&$variables) {


andybroomfield:
Would localgov_bases .theme file be a more suitable place for this? Presumably this only effects themes that sub theme from localgov_base (which might be installable without localgov_core).

Noting the discussion on https://github.com/localgovdrupal/localgov_base/issues/600#issuecomment-2783657214 that we might want to move the components into localgov_core in which case I think the setting will also need to move.

JohnAlbin:
> Would localgov_bases .theme file be a more suitable place for this?

`localgov_core.module` is where [`localgov_core_template_preprocess_default_variables_alter()`](https://github.com/localgovdrupal/localgov_core/pull/281/files#diff-d46919fa677915efbec29a06fd96d89b9a07638577f13e123fc8593a5891d9f1R47-R60) is currently located. And that hook sets the `localgov_base_remove_css` and `localgov_base_remove_js` twig variables that the `localgov_base` base theme and sub-themes use. So those theme settings already depend on `localgov_core`.

And while the `hook_library_info_alter` added in this PR _can_ be put into a `.theme` file, `hook_template_preprocess_default_variables_alter` cannot go in a theme.

So it seemed to make more sense to put these two hooks next to each other in `localgov_core.module`.

markconroy (approved):
Works a treat.

PR #283, opened 2025-04-08
With the view "files" disabled, localgov_core cause an exception

### LINKED ISSUE #282: With the view "files" disabled, localgov_core cause an exception
Because of https://github.com/localgovdrupal/localgov_core/blob/2.x/localgov_core.links.task.yml, if the files view is disabled, accessing the admin/content/media generates a fatal error due to the route defined in localgov_core.links.task.yml but not available.

Symfony\Component\Routing\Exception\RouteNotFoundException: Route "view.files.page_1" does not exist.

Even if the view likely exists in most of the cases, this is not granted, I think the local task should be defined dynamically and not statically https://www.drupal.org/docs/drupal-apis/menu-api/providing-module-defined-local-tasks#dynamic-local-task-generation

marcopagliarulo:
## What does this change?
Fixes #282

## How to test

1.  Disable the view files
2.  Open the path admin/content/media
3.  Notice that the page is loaded properly without error and the tab files is not available
4.  Re-enable the view files
5.  Open the path admin/content/media
3.  Notice that the page is loaded properly and the tab files is now available

The media view is accessible even when the files view is not available

### DIFF
--- a/localgov_core.links.task.yml
+++ b/localgov_core.links.task.yml
@@ -1,5 +1,3 @@
 localgov_core.files:
-  title: 'Files'
-  parent_id: entity.media.collection
-  route_name: view.files.page_1
+  deriver: 'Drupal\localgov_core\Plugin\Derivative\FilesLocalTasks'
   weight: 100
--- a/src/Plugin/Derivative/FilesLocalTasks.php
+++ b/src/Plugin/Derivative/FilesLocalTasks.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Drupal\localgov_core\Plugin\Derivative;
+
+use Drupal\Component\Plugin\Derivative\DeriverBase;
+use Drupal\Core\Routing\RouteProviderInterface;
+use Drupal\Core\Plugin\Discovery\ContainerDeriverInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides local task for files view under entity media collection.
+ */
+class FilesLocalTasks extends DeriverBase implements ContainerDeriverInterface {
+
+  /**
+   * The route provider.
+   *
+   * @var \Drupal\Core\Routing\RouteProviderInterface
+   */
+  protected $routeProvider;
+
+  /**
+   * Constructs a \Drupal\localgov_core\Plugin\Derivative\FilesLocalTasks instance.
+   *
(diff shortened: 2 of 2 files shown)


ekes:
As this is something that can be switched on and off-again, perhaps it would be better to do it in `hook_menu_local_tasks_alter()`?

marcopagliarulo:
I don't think so, this solution works even when switching on and off the view. I would avoid as much as possible procedural code

ekes:
> I don't think so, this solution works even when switching on and off the view.

OK then we should probably make a test for it.

> I would avoid as much as possible procedural code

https://www.drupal.org/node/3442349

marcopagliarulo:
@ekes thanks for the feedback, I'll try to reserve some time to it the coming days

ekes:
I was about to make a quick test, but then see that #221 was supposed to move the files view in the menu. At least on my installs at the moment I get the menu item under content _and_ the tab under media.

Found out this is because "This link is provided by the Admin Toolbar Extra Tools module. The title and path cannot be edited." I think we rather want to see if we can do something about that.
https://git.drupalcode.org/project/admin_toolbar/-/blob/5a0979ac832fddf3b2dc95da4044ebe4c0eecae0/admin_toolbar_tools/src/Plugin/Derivative/ExtraLinks.php#L718

So _thinking_ _aloud_ we want to get in later than that and change or remove that item and insert ours. I'm going to think the code on this PR could do it if we can be certain to come in later in the derivatives loop, or assuming the alter hook comes in later - that.

ekes:
> So thinking aloud we want to get in later than that and change or remove that item and insert ours. I'm going to think the code on this PR could do it if we can be certain to come in later in the derivatives loop, or assuming the alter hook comes in later - that.

OK so we can't remove the one from `admin_toolbar_tools` without implementing alter. So we unset that there. We can just leave the code in the MR here for adding in our own, even if it leaves menu alteration being done in two locations.

finnlewis:
I think we can close this one, @ekes re-worked the changes on see https://github.com/localgovdrupal/localgov_core/pull/297

Thanks @marcopagliarulo !

ekes:
Yes I think it didn't get auto closed because the commit was squashed.
But the individual commits in this branch are in the merged branch.
So thanks again @marcopagliarulo

PR #289, opened 2025-04-29
Update filter orders so image alignment works

### LINKED ISSUE #272: Text not wrapping around images
We have an issue with images on our website. The text just refuses to wrap around them! It appears to be behaving fine in the editing screen (see screenshot) but once published the text falls above/below the image. Example page here: https://www.tamworth.gov.uk/council/do-it-online 

Image

markconroy:
Closes #272 

## What does this change?

Moves the 'Align Images' filter to below "Correct HTML" so that text can wrap around images.

### DIFF
--- a/modules/localgov_media/config/optional/filter.format.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/filter.format.wysiwyg.yml
@@ -9,7 +9,6 @@ dependencies:
   module:
     - editor
     - linkit
-    - localgov_publications
     - media
 name: WYSIWYG
 format: wysiwyg
--- a/modules/localgov_media/config/optional/filter.format.wysiwyg.yml
+++ b/modules/localgov_media/config/optional/filter.format.wysiwyg.yml
@@ -9,6 +9,7 @@ dependencies:
   module:
     - editor
     - linkit
+    - localgov_publications
     - media
 name: WYSIWYG
 format: wysiwyg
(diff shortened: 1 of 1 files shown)


modules/localgov_media/config/optional/filter.format.wysiwyg.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
   module:
     - editor
     - linkit
+    - localgov_publications
```
ekes: Not sure it needs, or wants this dependency.

modules/localgov_media/config/optional/filter.format.wysiwyg.yml
```diff
@@ -18,58 +19,64 @@ filters:
...
+    weight: -39
     settings: {  }
   filter_htmlcorrector:
     id: filter_htmlcorrector
     provider: filter
     status: true
-    weight: -48
+    weight: -49
+    settings: {  }
+  filter_image_lazy_load:
```
ekes, line 68: This is a new filter.

So we're not just reordeing the filters here.

ekes (changes requested):

modules/localgov_media/config/optional/filter.format.wysiwyg.yml
markconroy: That lazy load is set by default when we install localgov. Should we leave it or remove it?

ekes: This is from Drupal Core? Doesn't need (any other) enabling? Does want to be configured on here (by default)?

Then I'd say we *explicitly* add it here then.

markconroy: Can we mark this thread resolved so? And approve the PR?

modules/localgov_media/config/optional/filter.format.wysiwyg.yml
markconroy: Removed

stephen-cox (approved):
Looking good to me now.

I did wonder about whether this needs an update hook, but I think it's highly likely people will have configured this text format themselves and it's a simple fix to apply manually if needed.

Issue #298, opened 2025-05-13
Show usage in the Media view


willguv:
Is it possible to show usage on this page?

/admin/content/media

Mentioned at the images content group session

As a content designer
I want to see at a glance the media that isn't be used
So I can delete it from the site

finnlewis:
@ekes wonders if this could be done with a view column and https://git.drupalcode.org/project/media_entity_usage

ekes:
Quick look, Media Entity Usage: 
 * Does not have a Drupal 11 version.
 * Does not count any other entities referencing that nodes and paragraphs (so you could have a taxonomy term with an image for example).

willguv:
We're blocked on this because the module is not Drupal 11 yet

ekes:
Oh, in addition only counting nodes and paragraphs: I don't think this is going to count linkit media usage either. We just introduced that?

ekes:
Ah that's better. We've got - now I remember the discussion about it going in!

https://www.drupal.org/project/entity_usage

which _does_ count all of the above, and have a D11 release.
And has some views integration. Lemme look.

willguv:
@ekes that's brilliant. If this is doable, I'd really like this in and usable by the end of May

ekes:
Transfered to localgov_core as that's where localgov_media is at the moment. This override is going to be like the one to views.view.files.yml (see my comment about finding a better home for that https://github.com/localgovdrupal/localgov_core/pull/297#issuecomment-2876379702 )

ekes:
PR https://github.com/localgovdrupal/localgov_core/pull/299

willguv:
Wow, fast! Thanks @ekes 

As I can't install this right now, does this follow the same UX as paragraphs?

/admin/content/paragraphs

A column with "Used" as a heading, immediately to the left of the changed/ updated column?

ekes:
Screenshot on PR. 'Usage count' is the title - can change that, if 'Used' is clear, it's shorter. Presently it's the last column before the dropdown - it could be between status and updated. Then the number is a link to the /media/{id}/edit/usage page.

willguv:
@ekes shall we follow same pattern as paragraphs?

- Heading is 'Used'
- Position is 3rd from right

<img width="1561" alt="Image" src="https://github.com/user-attachments/assets/10b4ad5f-b62b-4bbd-ac66-a3f48e14dd97" />

Do we allow sorting by used? Think we should

ekes:
[...]

PR #297, opened 2025-05-13
 With the view "files" disabled, localgov_core cause an exception #283 

ekes:
Push of further developed branch from #283

PR #299, opened 2025-05-15
Feature #298 media list usage display

ekes:
Adds an alternative admin view for admin media page including a usage column.

Update hook will replace view if it's not been changed (and entity_usage is installed).

Manual tests:
* Install fresh: /admin/content/media should have a usage count column;
* Install on existing site: if the /admin/content/media view has not been changed it will be update; if it has been changed it should not be updated; if entity_usage is uninstalled it should not be updated.

Note about automated tests: This suffers from the same issues as https://github.com/localgovdrupal/localgov_core/pull/297#issuecomment-2876379702 We should move these functionalities elsewhere at some point and work out how to make reliable tests.

### DIFF
--- a/modules/localgov_media/config/optional/views.view.media.yml
+++ b/modules/localgov_media/config/optional/views.view.media.yml
@@ -389,6 +389,116 @@ display:
           multi_type: separator
           separator: ', '
           field_api_classes: false
+        edit_media:
+          id: edit_media
+          table: media
+          field: edit_media
+          relationship: none
+          group_type: group
+          admin_label: ''
+          entity_type: media
+          plugin_id: entity_link_edit
+          label: ''
+          exclude: true
+          alter:
+            alter_text: false
+            text: ''
+            make_link: false
+            path: ''
+            absolute: false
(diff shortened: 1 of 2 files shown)


ekes:
image

ekes:
image

markconroy (approved):
Works a treat on a new install.

tonypaulbarker:
This is a nice enhancement. For follow up, wondering if it would work to display a link (or the list of items) on the media entity edit pane so people could see the usage at the point of editing.

markconroy:
@tonypaulbarker Good idea. Wanna create a new issue for that and we can work on it then?

tonypaulbarker:
Follow up issue created at https://github.com/localgovdrupal/localgov_core/issues/300

willguv:
@ekes @tonypaulbarker I'd like to write about this in the next product newsletter (goes out first week of the month)

How do users get this please?

- ensure entity_usage is installed (think this should be in the profile, is it?)
- update localgov_core?

PR #303, opened 2025-05-23
Add a dependency on core to media view

### LINKED ISSUE #302: Test failure in profile due to #298
Since the merging and releasing of #298 there's a test failure in the profile due to missing schema for localgov_page_header_display_extender. This is due to the new media view having a dependency on localgov_core (where the schema is defined).

There are 2 options to solve this:

1) Add a dependency in localgov_media to localgov_core.

2) Remove the config referring to localgov_page_header_display_extender from the new media view.

https://github.com/localgovdrupal/localgov_core/blob/15d908218a176b09e3dc0884ce14c6809bd71474/modules/localgov_media/config/optional/views.view.media.yml#L1048-L1050

I'll create a PR for option 1 for discussion.

stephen-cox:
Fixes #302 

## What does this change?

~~Adds a dependency on localgov_core to localgov_media. See #302 for an alternative approach.~~

Adds a dependency on localgov_core to media view in localgov_media.

## How to test

The testLocalGovDrupalProfile PHPUnit test in the profile passes.

~~Is there a possibility that anyone is using localgov_media without localgov_core?~~

### DIFF
--- a/modules/localgov_media/config/optional/views.view.media.yml
+++ b/modules/localgov_media/config/optional/views.view.media.yml
@@ -6,6 +6,7 @@ dependencies:
   module:
     - entity_usage
     - image
+    - localgov_core
     - media
     - user
 id: media
--- a/modules/localgov_media/localgov_media.info.yml
+++ b/modules/localgov_media/localgov_media.info.yml
@@ -12,3 +12,4 @@ dependencies:
   - drupal:responsive_image
   - image_widget_crop:image_widget_crop
   - linkit:linkit
+  - localgov_core:localgov_core
--- a/modules/localgov_media/localgov_media.info.yml
+++ b/modules/localgov_media/localgov_media.info.yml
@@ -12,4 +12,3 @@ dependencies:
   - drupal:responsive_image
   - image_widget_crop:image_widget_crop
   - linkit:linkit
-  - localgov_core:localgov_core


tonypaulbarker:
I'm not sure if it's possible to add localgov_media without localgov_core in theory but it seems unlikely on an LGD site. It seems worthwhile anyway for keeping things minimal and tidy to explore whether localgov_page_header_display_extender is really needed, removable or adjustable.

stephen-cox:
For further context localgov_page_header_display_extender was added with https://github.com/localgovdrupal/localgov_core/pull/257

This is break loads of tests at the moment (see https://github.com/localgovdrupal/localgov_project/actions/runs/15209084783/job/42778869156?pr=161) so we need a solution for this ASAP.

ekes:
Think this highlights the dependencies in media (already complex, and it has been mentioned it should be split apart several times), and that we didn't know which module #257 should really go into: core was a best for now option.

Also this is a failing test on optional configuration. So it's not a hard dependency. Only the test just needs it if that view is being enabled. The view should not be enabled, the optional config should not be imported, if the dependency isn't there outside of a test environment.

finnlewis (approved):
Makes sense for now, localgov_media is a sub module of localgov_core for now.

Shall we look at breaking it out and teasing apart the dependencies on a separate issue?

ekes:
> Makes sense for now, localgov_media is a sub module of localgov_core for now.
> Shall we look at breaking it out and teasing apart the dependencies on a separate issue?

I don't understand why an optional config item needs a module hard dependency. The point of optional config is it's optional and only installed if the dependencies are available.

So my concern now is that, and that 'for now' has been our answer for all the previous cases! Including where should 257 go.

stephen-cox:
The reason optional config it causing this is the media view doesn't declare a dependency on localgov_core, when it clearly should.

https://github.com/localgovdrupal/localgov_core/blob/2.x/modules/localgov_media/config/optional/views.view.media.yml#L6-L10

So moving the dependency to the view config should fix this. I will try it.

stephen-cox:
> So moving the dependency to the view config should fix this. I will try it.

Seems to work :) Will rework this PR

PR #308, opened 2025-06-25
Media module container_rebuild_required key for D11.2

### LINKED ISSUE #307: Add container_rebuild_required key to localgov_media to resolve Drupal 11.2 config dependencies
container_rebuild_required defaults to false in 11.2. This causes some config dependency issues https://github.com/localgovdrupal/localgov_paragraphs/issues/247
and is a similar treatment to https://github.com/localgovdrupal/localgov_topics/pull/35 

https://www.drupal.org/project/drupal/issues/3529670 

Alex said "try adding  `container_rebuild_required: true` to modules/localgov_media/localgov_media.info.yml - this should fix this."

### How to test

Currently failing tests related to media config should pass for Drupal 11.2 
Should not cause regressions for previous versions
The project should install for Drupal 11.2

tonypaulbarker:
## What does this change?

Fixes https://github.com/localgovdrupal/localgov_core/issues/307
Introduces container_rebuild_required key for the media module for 11.2+ support

## How to test

- Tests should no longer complain about the media config, e.g. https://github.com/localgovdrupal/localgov_paragraphs/issues/247 

- The project should install on 11.2
- Should not be regressions when installing for older versions of Drupal

We can move closer to Drupal 11.2 support. 

The key breaks something in older versions
We enter a game of 'whack-a-mole' where we move on just one step and see another failure.

### DIFF
--- a/modules/localgov_media/localgov_media.info.yml
+++ b/modules/localgov_media/localgov_media.info.yml
@@ -3,6 +3,7 @@ description: LocalGov Media configuration.
 package: LocalGov Drupal
 type: module
 core_version_requirement: ^10 || ^11
+container_rebuild_required: true
 
 dependencies:
   - drupal:ckeditor5


millnut:
@tonypaulbarker can this come out of draft and get included in a release so I can run a full D11.2 test

tonypaulbarker:
@millnut @stephen-cox  some good news, this does appar to nudge us along. I've marked it ready for review.

Until this is reviewed and merged I think a command like this can be used for testing:
composer require "localgovdrupal/localgov_core:dev-fix/2.x/307-container-rebuild-localgov-media as 2.x-dev"

I had to take a few additional steps noted here:

Mariadb in ddev to 10.6 https://github.com/localgovdrupal/localgov_project/issues/197
Bump brianium/paratest to 7

I saw the permissions error that this may fix https://github.com/localgovdrupal/localgov_project/issues/197 
I think linting tools are needing updating https://github.com/localgovdrupal/localgov_base/issues/701 

Also to note that for me Drupal 10.5 did not complain about the new key.

millnut:
Hi @tonypaulbarker yep those changes (mariadb, paratest, etc...) are already in the [3.2.x](https://github.com/localgovdrupal/localgov_project/tree/3.2.x) branch of the project which contains all the other Drupal 11.x work

stephen-cox:
@tonypaulbarker I tried adding `container_rebuild_required: true` to all the LGD modules, but the tests that require config in the media module still fail. It would be good if someone else can config this.

Locally, I edited all the info.yml files to add the new key and then ran the test in the localgov_homepage_paragraphs module. It still fails with a missing dependency on media.type.image. The config for this is in localgov_media and moving it from optional to install fixes the test.

tonypaulbarker:
Thanks for the information @stephen-cox .

Having analysed some more I think media dependencies are doing the trick on a 'full' install but the unit test requires some limited things in isolation at the moment so fails. Only moving 'types' might get those tests to pass but moving everything out of optional is probably the right move because in practice the optional dependencies are always satisfied.

finnlewis:
Closing as duplicate of #316

PR #315, opened 2025-08-01
fix: add missing dependencies for localgov_media

millnut:
## What does this change?

Adds missing dependencies based on the config

### DIFF
--- a/modules/localgov_media/localgov_media.info.yml
+++ b/modules/localgov_media/localgov_media.info.yml
@@ -9,6 +9,10 @@ dependencies:
   - drupal:editor
   - drupal:media_library
   - drupal:media_library_edit
+  - drupal:path
   - drupal:responsive_image
+  - entity_usage:entity_usage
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Looks good, thanks @millnut

finnlewis:
@millnut ... one sec - @stephen-cox just pointed out that paragraphs is only in require-dev in composer.json.

https://github.com/localgovdrupal/localgov_core/blob/2.x/composer.json#L19

Is this holding anything up or helping in a big way?

tonypaulbarker (changes requested):
@millnut marking as needing changes to address paragraphs:

> @millnut ... one sec - @stephen-cox just pointed out that paragraphs is only in require-dev in composer.json.

> https://github.com/localgovdrupal/localgov_core/blob/2.x/composer.json#L19

millnut:
@stephen-cox are you happy for me to move paragraphs from a dev dependency to a dependency?

PR #320, opened 2025-09-23
Add media test config module for use in tests 

stephen-cox:
## What does this change?

Tests are failing on Drupal 11.2 due to changes to how modules are installed which breaks tests due to how LGD media config is installed.

This adds a module that duplicates the config in localgov_media, but moves the config from optional to install. This can then be included in tests to ensure the necessary config is in place during testing.

For an example of how tests will need to be changed after this is merged and released see: https://github.com/localgovdrupal/localgov_events/pull/203/files

## How to test

Tests that rely on LGD media config should pass when run with Drupal 11.2


tonypaulbarker:
@stephen-cox the only issue I foresee here is if some other module depends on installing localgov_media and the test tries to install both copies. Do we need to account for that scenario in any of the modules?

stephen-cox:
> the only issue I foresee here is if some other module depends on installing localgov_media and the test tries to install both copies. Do we need to account for that scenario in any of the modules?

We should be fine. All the config in localgov_media is in optional so if there's an attempt to installing this it will be skipped because it already exists.

finnlewis:
Discussed on Merge Tuesday and we're happy to merge to help fix the tests for Drupal 11 :)

finnlewis:
> ## What does this change?
> Tests are failing on Drupal 11.2 due to changes to how modules are installed which breaks tests due to how LGD media config is installed.
> 
> This adds a module that duplicates the config in localgov_media, but moves the config from optional to install. This can then be included in tests to ensure the necessary config is in place during testing.
> 
> For an example of how tests will need to be changed after this is merged and released see: https://github.com/localgovdrupal/localgov_events/pull/203/files
> 
> ## How to test
> Tests that rely on LGD media config should pass when run with Drupal 11.2

On drupal.org, we have failing tests on localgov_blogs.

1) Drupal\Tests\localgov_blogs\Functional\BlogCreationTest::testBlogChannelSelection
Unable to install modules: module 'localgov_media_test_config' is missing its dependency module paragraphs_library.

https://git.drupalcode.org/project/localgov_blogs/-/jobs/6753956

I think this is related to this solution, so trying it!

But do we need to include dependencies on other modules like paragraphs_library in localgov_media_test_config or in localgov_blogs ?

tonypaulbarker:
@finnlewis I did have a look at the failing test and since one module’s dependencies affects another I will look to open issues / PRs against GitHub for missing dependencies along with some notes.

PR #324, opened 2025-10-06
adds Microsites Base to themes array

### LINKED ISSUE #323: Add localgov_microsites_base to themes array for block placement
If we add `localgov_microsites_base` to the themes array, then we can remove custom block placement config from the submodules of `localgov_microsites_group`.

markconroy:
Closes #323

## What does this change?

Adds `localgov_microsites_base` to themes array for block installer, so we don't have to add that config in `localgov_microsites_group` submodules.

---


finnlewis (approved):
@tonypaulbarker seems sensible and harmless.

Looks good to us!

PR #332, opened 2025-11-03
Try a recipe for test media module

### LINKED ISSUE #330: Test media config fails on Drupal 10
Test config module was introduced in https://github.com/localgovdrupal/localgov_core/pull/320 
LocalGov Blogs tests report the media config is already present.

tonypaulbarker:
Closes https://github.com/localgovdrupal/localgov_core/issues/330

Success criteria: 

Tests like those in localgov_blogs should be able to install this in both Drupal 10 and Drupal 11 to get their config dependencies.


tonypaulbarker:
This MR is a dead end.

finnlewis:
Hey @tonypaulbarker , why is it a dead end, what did it reveal?

tonypaulbarker:
@finnlewis the same timing problem occurs here as with the original optional directory in the localgov_media module so we see the same errors. 

1. Dirty workaround - check for Drupal version and conditionally install the config if greater than or equal to 11.2

2. Debug the timing in lots of detail and try rebuilding containers in the tests another way

Try this sort of thing to work around the container rebuild problem:
`function localgov_media_install() {
  if (version_compare(\Drupal::VERSION, '11.2', '>=')) {
    if (drupal_valid_test_ua() {
      \Drupal::service('kernel')->invalidateContainer();
    }
  }
}`  

Try to rebuild at certain points in the test
`$this->container = $this->kernel->rebuildContainer();`

Try controlling the timing of installing things in the test 
`\Drupal::service('module_installer')->install(['localgov_media']);` 

3. Fix up the media module to really install things another way than hook and optional directory.

andybroomfield:
Something like [Configuration Rewrite](https://www.drupal.org/project/config_rewrite) be useful here?

If I remember correctly the underlying issue is we install media config that already exists in standard profile. Could we replace it if it exists (or skip)?

PR #335, opened 2025-12-01
adds unpublished items to linkit

### LINKED ISSUE #334: Unpublished nodes should be able to be referenced through linkit
### Problem statement

With current settings, only published nodes show up in the linkit list. This means that editors drafting a website section are blocked from linking to other unpublished pages in their section until publication time.

### Actions:

Check [the box (insert details)] to enable referencing unpublished nodes in the widget.
Update hook to turn on for existing sites. 
Consider a theming change for UX (see below)

### UX consideration

The UX for seeing that a node is unpublished is that the item is faintly pink in the dialog.

Image

Should we introduce a theming change to make it clearer in the list that an item is unpublished? We should think about not using colour to convey important information here.

### Risks

If editors publish a page with links to unpublished content, end users without permissions may follow a link to an access denied page.

markconroy:
Closes #334

## What does this change?

- Allows unpublished items to be linkit suggestions
- Adds "Unpublished: " before the title if it's unpublished.

## How to test
- Checkout this branch
- Clear your cache
- Make sure 'unpublished items' are allowed in the linkit matcher

Content editors can create whole site sections without them being published.

## Images

Screenshot 2025-12-01 at 12 15 24

----

### DIFF
--- a/modules/localgov_media/config/optional/linkit.linkit_profile.default.yml
+++ b/modules/localgov_media/config/optional/linkit.linkit_profile.default.yml
@@ -2,6 +2,7 @@ langcode: en
 status: true
 dependencies:
   module:
+    - localgov_core
     - node
     - paragraphs_library
 label: Default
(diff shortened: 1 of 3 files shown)


tonypaulbarker (approved):
This works well, we need to merge on drupal.org gitlab

